PR #1, opened 2019-11-04
A very simple first version of snare.

ltratt:
This is a very simple GitHub webhooks runner. Hopefully the README is relatively clear as to what it does! This could be improved in so many ways, but I suspect this (or something roughly like this) is adequate for our purposes for the time being. It has survived basic testing, at least!

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,1 +1,74 @@
 # snare
+
+`snare` is a simple GitHub webhooks runner. When a request comes in, it runs an
+arbitrary program for that repository informing it of the event -- that program
+can then perform whatever actions it wants.
+
+
+## Basic usage
+
+`snare` requires three command-line arguments to be specified:
+
+```
+Usage: snare -p <port> -r <repo-programs-dir> -s <secrets-path>
+```
+
+where:
+
+ * `<port>` is a port number (e.g. 4567).
+ * `<repo-programs-dir>` is the directory where the per-repo programs are
+   stored. For a repository `repo` owned by `user` the command
+   `<repo-programs-dir>/<user>/<repo> <event> <path to GitHub JSON>` will be
+   run.
+ * `<secrets-path>` is the file containing the GitHub secret which guarantees
+   that hooks are coming from your GitHub repository and not a malfeasant. Note
+   that leading and trailing whitespace in this file is ignored.
+
+The most important part of this are the per-repo programs. For example,
+`snare`'s GitHub repository is
+[`https://github.com/softdevteam/snare`](https://github.com/softdevteam/snare).
+If we set up a web hook up for that repository that notifies us of pull request
+events, then the command:
+
+```sh
+<repo-programs-dir>/softdevteam/snare pull_request /path/to/json
+```
+
+will be executed, where: `pull_request` is the name of the GitHub event; and
+`/path/to/json` is a path to a file containing the complete GitHub JSON for
+that event. The `softdevteam_snare` program can then execute whatever it wants.
+In order to work out precisely what event has happened, you will need to read
+(GitHub's webhooks documentation)[https://developer.github.com/webhooks/].
+
+
+## Example repository program
+
+If we want to only execute commands when a pull request is merged, your
+per-repo program might start as follows:
+
+```sh
+#! /bin/sh
+
+# Ignore everything except pull request events
+if [ $1 != "pull_request" ]; then
+    exit 0
+fi
+
+# Ignore pull request events that aren't closing a pull request
+if [ "X`jq .action $2 | tr -d '\"'`" != "Xclosed" ]; then
+    exit 0
+fi
+
+# Ignore close events unless they merged changes in
+if [ "X`jq .pull_request.merged $2 | tr -d '\"'`" != "Xtrue" ]; then
+    exit 0
+fi
+```
+
+where (`jq`)[https://stedolan.github.io/jq/] is a command-line JSON processor.
+If all three of those `if` statements succeed, then we know that a pull request
+has been merged. As this suggests, some GitHub events are slightly trickier
+than others to process and writing the above in shell script doesn't make it
+particularly easy to see the core logic. However, users can equally well write
+such programs in other languages if they prefer (i.e. you don't need to write
+shell scripts for this if you don't want to).
--- a/snare
+++ b/snare
@@ -1,0 +1,75 @@
+#! /usr/bin/env python3.7
+
+# Copyright (c) 2019 King's College London, created by the Software Development
+# Team <http://soft-dev.org/>
+#
+# Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+# http://www.apache.org/licenses/LICENSE-2.0>, or the MIT license <LICENSE-MIT
+# or http://opensource.org/licenses/MIT>, or the UPL-1.0 license
+# <http://opensource.org/licenses/UPL> at your option. This file may not be
+# copied, modified, or distributed except according to those terms.
+
+import getopt, hashlib, hmac, json, os, subprocess, sys, tempfile
+from http.server import ThreadingHTTPServer, BaseHTTPRequestHandler
+from urllib.parse import unquote
+
+class HookHandler(BaseHTTPRequestHandler):
+    def do_POST(self):
+        data = self.rfile.read(int(self.headers["Content-Length"]))
+        if not self._check_secret(data):
+            self.send_response(401)
+            return
+
+        data_str = unquote(data.decode("utf-8")[8:])
+        pl = json.loads(data_str)
+        owner = pl["repository"]["owner"]["login"].lower()
+        repo = pl["repository"]["name"].lower()
+
+        path = os.path.join(repo_scripts_dir, owner, repo)
+        if os.path.exists(path):
+            with tempfile.NamedTemporaryFile() as jsonf:
+                jsonf.write(data_str.encode())
+                event = self.headers["X-GitHub-Event"]
+                rtn = subprocess.run([path, event, jsonf.name])
+            if rtn.returncode == 0:
+                self.send_response(200)
+            else:
+                self.send_response(500)
+        else:
+            self.send_response(501)
+
+    def _check_secret(self, data):
+        sha, sig = self.headers["X-Hub-Signature"].split("=")
+        mac = hmac.new(secret.encode(), msg=data, digestmod=hashlib.sha1)
+        return hmac.compare_digest(mac.hexdigest(), sig)
+
+def usage(code):
+    sys.stderr.write("Usage: snare -p <port> -r <repo-programs-dir> -s <secrets-path>\n")
+    sys.exit(code)
+
+if __name__ == '__main__':
+    port = None
+    repo_scripts_dir = None
+    secrets_path = None
+    opts, args = getopt.getopt(sys.argv[1 : ], "hp:r:s:")
+    for opt in opts:
+        if opt[0] == "-h":
+            usage(0)
+        elif opt[0] == "-p":
+            port = int(opt[1])
+        elif opt[0] == "-r":
+            repo_scripts_dir = os.path.abspath(opt[1])
+        elif opt[0] == "-s":
+            secrets_path = opt[1]
+    if port is None or repo_scripts_dir is None or secrets_path is None or len(args) > 0:
+        usage(1)
+
+    with open(secrets_path) as f:
+        secret = f.read().strip()
+
+    while True:
+        try:
+            server = ThreadingHTTPServer(('', port), HookHandler)
+            server.serve_forever()
+        except Exception:
+            pass


snare
```diff
@@ -0,0 +1,68 @@
...
+        repo = pl["repository"]["name"].lower()
+
+        path = os.path.join(repo_scripts_dir, "%s_%s" % (owner, repo))
+        if os.path.exists(path):
+            with tempfile.NamedTemporaryFile() as jsonf:
+                jsonf.write(data_str.encode())
+                event = self.headers["X-GitHub-Event"]
+                subprocess.run([path, event, jsonf.name])
+
+        self.send_response(200)
```
ptersilie: We may want to allow users to return their own status codes and messages here. For example, they may want to return something like `400, "Not yet implemented"`, which will then show up in their GitHub log (e.g. https://github.com/<user>/settings/hooks/<hookid>), making it easier to debug their scripts.

ltratt: I'm not really sure what to do with the return codes because there's no guarantee that the script completes in time to return a code (e.g. yksom's takes long enough to complete that GitHub assumes that things have timed out). That said, returning 400 if there isn't a script is probably a good idea

ltratt: See if https://github.com/softdevteam/snare/pull/1/commits/78a3e3ad97a0973fcb76819e3f9324f6ab5875d5 makes things a bit better. It's not configurable, but it's a little better I think.

ptersilie: > there's no guarantee that the script completes in time to return a code

That's fine, in those cases we are already not returning `200`, but in the cases we do see a return code I think it would still be nice if users could define it themselves along with a message. When I did the plugin for Mattermost it was very useful to see which hooks I haven't implemented yet as they would be highlighted in the GitHub log.

ltratt: Maybe we can do customisation in the future? I'm not sure how to handle it sensibly at the moment, and we don't really need it at the moment either.

ptersilie: I'm happy to make this a future feature, I just wanted to raise it as from my own experience working with GitHub webhooks this can be very useful.

ptersilie:
Please squash.

ltratt:
I'd just like to quickly double check that this does what @vext01 would like us to do before squashing.

README.md
```diff
@@ -1 +1,66 @@
...
+arbitrary program for that repository informing it of the event -- that program
+can then perform whatever actions it wants.
+
+
+## Basic usage
+
+`snare` requires three command-line arguments to be specified:
+
+```
+Usage: snare -p <port> -r <repo scripts dir> -s <secret>
```
vext01: The secret is going to appear in the process table if it is an argument to the command. Better to read from a file?

README.md
```diff
@@ -1 +1,66 @@
...
+
+where:
+
+ * `<port>` is a port number (e.g. 4567)
+ * `<repo scripts dir>` is the directory where programs for each repository
+   targeted by the webhook are stored. For a repository `repo` owned by `user`
+   the command `<repo scripts dir>/<user>_<repo> <event> <path to GitHub JSON>`
+   will be run.
+ * `<secret>` is the GitHub secret which guarantees that hooks are coming from
+   your GitHub repository and not a malfeasant.
```
vext01: I had to look up the word `malfeasant`. Seems to have political implications? I wonder if simpler language would be better?

README.md
```diff
@@ -1 +1,66 @@
...
+
+The most important part of this are the per-repo programs. For example,
+`snare`'s GitHub repository is
+[`https://github.com/softdevteam/snare`](https://github.com/softdevteam/snare):
+if we set up a web hook up that notifies us of when pull request events have
+been made then the command `<repo scripts dir>/softdevteam_snare pull_request
+/path/to/json` will be executed where `pull_request` is the name of the GitHub
+event and `/path/to/JSON` is a path to a file containing the complete GitHub
```
vext01: Is it `json` or `JSON`?

README.md
vext01: I think `<repo-scripts-dir>` is easier to read (with hyphens), but I don't feel strongly.

README.md
```diff
@@ -1 +1,66 @@
...
+
+ * `<port>` is a port number (e.g. 4567)
+ * `<repo scripts dir>` is the directory where programs for each repository
+   targeted by the webhook are stored. For a repository `repo` owned by `user`
+   the command `<repo scripts dir>/<user>_<repo> <event> <path to GitHub JSON>`
+   will be run.
+ * `<secret>` is the GitHub secret which guarantees that hooks are coming from
+   your GitHub repository and not a malfeasant.
+
+The most important part of this are the per-repo programs. For example,
```
vext01: This is the first mention of "per-repo programs". Is it the same as "repo scripts dir"?

README.md
```diff
@@ -1 +1,66 @@
...
+`snare`'s GitHub repository is
+[`https://github.com/softdevteam/snare`](https://github.com/softdevteam/snare):
+if we set up a web hook up that notifies us of when pull request events have
+been made then the command `<repo scripts dir>/softdevteam_snare pull_request
+/path/to/json` will be executed where `pull_request` is the name of the GitHub
+event and `/path/to/JSON` is a path to a file containing the complete GitHub
+JSON for that event. The `softdevteam_snare` program can then execute whatever
+it wants. In order to work out precisely what event has happened, you will
+need to read (GitHub's webhooks
+documentation)[https://developer.github.com/webhooks/].
```
vext01: I don't understand this paragraph.

> the command `<repo scripts dir>/softdevteam_snare pull_request /path/to/json` will be executed

Sound like you execute a JSON file? That can't be right.

> `/path/to/JSON` is a path to a file containing the complete GitHub JSON for that event.

So you write what github gives you into JSON file? That can't be right either. I must have misunderstood.

vext01: > the command <repo scripts dir>/softdevteam_snare pull_request /path/to/json will be executed

Wait, I just noticed the spaces. So you do in fact write the JSON to a file, and pass it to a program the user has specified?

ptersilie: The script you run will need access to the entire JSON that GitHub sends in order to make decisions based on it. For example, let's say we execute `pr_event.py` whenever a PullRequest event occurs. At this point we don't know what kind of PullRequest event it is. It could be `assigned`, `closed`, `merged`, etc. The script will thus need access to the entire JSON data, which is rather large, in order to do the right thing.

vext01: OK, that's fine. I just wanted to check I had understood.

README.md
```diff
@@ -1 +1,66 @@
...
+if [ "X`jq .action $2 | tr -d '\"'`" != "Xclosed" ]; then
+    exit 0
+fi
+
+if [ "X`jq .pull_request.merged $2 | tr -d '\"'`" != "Xtrue" ]; then
+    exit 0
+fi
+```
+
+where `jq` is a command-line JSON processor. If all three of those `if`
```
vext01: Should we mention that `jq` is not just any JSON processor here?

README.md
```diff
@@ -1 +1,66 @@
...
+
+if [ $1 != "pull_request" ]; then
+    exit 0
+fi
+
+if [ "X`jq .action $2 | tr -d '\"'`" != "Xclosed" ]; then
+    exit 0
+fi
+
+if [ "X`jq .pull_request.merged $2 | tr -d '\"'`" != "Xtrue" ]; then
```
vext01: I think these checks could be one-liners?

```
[ "X`jq .pull_request.merged $2 | tr -d '\"'`" != "Xtrue" ] && exit 0
```

README.md
ltratt: > The secret is going to appear in the process table if it is an argument to the command. Better to read from a file?

Good point.

README.md
ltratt: Good spot. In the path it should be `json.

README.md
ltratt: I guess they could. I've never tried writing them like that!

snare
```diff
@@ -0,0 +1,72 @@
...
+# Copyright (c) 2019 King's College London, created by the Software Development
+# Team <http://soft-dev.org/>
+#
+# Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+# http://www.apache.org/licenses/LICENSE-2.0>, or the MIT license <LICENSE-MIT
+# or http://opensource.org/licenses/MIT>, or the UPL-1.0 license
+# <http://opensource.org/licenses/UPL> at your option. This file may not be
+# copied, modified, or distributed except according to those terms.
+
+import getopt, hashlib, hmac, json, os, subprocess, sys, tempfile, traceback
```
vext01: flake8 found:

```
snare:12:1: F401 'traceback' imported but unused
```

It also found a load of code style nits (long lines, whitespace etc), but I guess you are not doing PEP8.

snare
```diff
@@ -0,0 +1,72 @@
+#! /usr/bin/env python3.7
+
+# Copyright (c) 2019 King's College London, created by the Software Development
+# Team <http://soft-dev.org/>
+#
+# Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
```
vext01: Is it me, or does it look odd to split `<...>` over multiple lines? It took me a while to parse.

(Not super-important)

ltratt: When it's that long, I don't think there's much choice.

snare
ltratt: Ah, yes, I had removed the need for `traceback`. Will remove the import.

snare
```diff
@@ -0,0 +1,72 @@
...
+from urllib.parse import unquote
+
+class HookHandler(BaseHTTPRequestHandler):
+    def do_POST(self):
+        data = self.rfile.read(int(self.headers["Content-Length"]))
+        if not self._check_secret(data):
+            self.send_response(401)
+            return
+
+        data_str = unquote(data.decode("utf-8")[8:])
```
vext01: utf-8 is default encoding, so you could kill the explicit char encoding.

https://docs.python.org/3/library/stdtypes.html?highlight=decode#bytes.decode
```
bytes.decode(encoding="utf-8", errors="strict")
```

Does the JSON github send you specify an encoding actually?

vext01: what's the `[8:]` for?

snare
```diff
@@ -0,0 +1,72 @@
...
+        pl = json.loads(data_str)
+        owner = pl["repository"]["owner"]["login"].lower()
+        repo = pl["repository"]["name"].lower()
+
+        path = os.path.join(repo_scripts_dir, "%s_%s" % (owner, repo))
```
vext01: I've started using f-strings in new Python3 code. You could make this line:
```
path = os.path.join(repo_scripts_dir, f"{owner}_{repo}")
```

snare
```diff
@@ -0,0 +1,72 @@
...
+        elif opt[0] == "-s":
+            secret = opt[1]
+    if port is None or repo_scripts_dir is None or secret is None or len(args) > 0:
+        usage(1)
+
+    while True:
+        try:
+            server = ThreadingHTTPServer(('', port), HookHandler)
+            server.serve_forever()
+        except Exception:
```
vext01: I recommend you don't catch such a generic exception. I used to do this in krun, and it hides bugs and makes debugging a nightmare.

README.md
vext01: Weather it's easier to read or not is another question of course! Up to you.

README.md
ltratt: It's a great word!

README.md
ltratt: I agree that tidying this up would be good.

README.md
ltratt: People can use whatever they want, but `jq` is one way of doing it. I've linked to the `jq` website so that interested parties can investigate it.

README.md
ltratt: They could... except in `dash` which barfs on this. Sigh.

snare
ltratt: The string starts with `payload=` which needs stripping.

snare
ltratt: I actually think it's clearer as it is!

snare
ltratt: In a server, you've got to keep serving no matter what, so there isn't much choice.

ltratt:
@vext01 See if https://github.com/softdevteam/snare/pull/1/commits/04fd4e7cc4be86d543bbb462eb5a270560d5cfba works for you.

README.md
```diff
@@ -10,57 +10,66 @@ can then perform whatever actions it wants.
...
  * `<port>` is a port number (e.g. 4567)
- * `<repo scripts dir>` is the directory where programs for each repository
-   targeted by the webhook are stored. For a repository `repo` owned by `user`
-   the command `<repo scripts dir>/<user>_<repo> <event> <path to GitHub JSON>`
-   will be run.
- * `<secret>` is the GitHub secret which guarantees that hooks are coming from
-   your GitHub repository and not a malfeasant.
+ * `<repo-scripts-dir>` is the directory where the per-repo programs are
+   stored. For a repository `repo` owned by `user` the command
+   `<repo-scripts-dir>/<user>_<repo> <event> <path to GitHub JSON>` will be
```
vext01: This is a nitpick, but I wonder if if `<repo-scripts-dir>/user/repo` would be better.

Why?
`<repo-scripts-dir>/simon_123_flibble`

Is this?
 - user: simon, repo: 123_flibble
 - user: simon_123, repo: flibble

If both of those repos existed, the path would clash.

ltratt: Good point.

README.md
vext01: Doh! Never mind.

snare
vext01: I understand that you want to keep going if there's some temporary network error, but some classes of exception you are going to want to know about.

Eg `IndexError`, `KeyError`, `ValueError`. These would all indicate bugs in our script, and your catch all would silence them. We'd never know that our hooks are not executing.

ltratt: Then we will crash, and we won't get a bug report either. We can't really win, so it's generally better for the server to keep running.

README.md
ltratt: Resolved in https://github.com/softdevteam/snare/pull/1/commits/0b0aa351cf4006c5e2c25d187ee1f46f22afef60

snare
vext01: don't you at least want to print the exception before continuing?

ltratt: It does.

vext01: Based on my experience with Krun, I think this decision will come back to haunt you. Ultimately it's up to you, since you are maintaining it ;)

Another option would be to notify someone (via email?) if X hooks have failed.

Come to think about it, wouldn't you want to know if any single hook failed? Otherwise our docs will be out of date. Do we need some kind of retry logic then?

ltratt: At the moment everything gets logged to stderr/stdout. I'm wary of doing anything more clever until I understand a bit better what the failure modes are.

vext01: OK. So you are going to manually check stderr/stdout to start with? Perhaps run in a tmux?

ltratt: Yes, in a tmux.

vext01: Sorry to bang on, but where do you print to stderr/stderr? It looks like you catch any exception and silently discard it.

Wouldn't you want:
```
except Exception as e:
    print(e)
```

?

ltratt: The HTTP server prints the exception.

vext01: Ah I see!

ltratt:
OK to squash?

snare
vext01: Reading the `serve_forever()` docs, it's not clear if the server will re-raise exceptions. But I take it you added the handler because it was?

It looks like it boils down to a call to:
https://docs.python.org/2/library/socketserver.html#SocketServer.BaseServer.handle_error

(I promise I will leave you alone soon)

ltratt: It works as is even when things go a bit wrong. Perfect? Probably not. Adequate? For now, yes.

vext01: Point taken.

vext01:
I'm OK with this. If @ptersilie is OK too, it's squash time.

ptersilie:
Yep!

vext01:
Please squash.

ltratt:
Squashed.

PR #4, opened 2019-11-08
Always run per-repo programs in a temporary directory.

ltratt:
It seems likely that nearly all per-repo programs will want to write something to disk (at the very least cloning a repository), so running them in a temporary directory (which is automatically cleaned up after the program exits) makes the programs a little less error prone and a little easier to write.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,9 @@ where:
  * `<repo-programs-dir>` is the directory where the per-repo programs are
    stored. For a repository `repo` owned by `user` the command
    `<repo-programs-dir>/<user>/<repo> <event> <path to GitHub JSON>` will be
-   run.
+   run. Note that per-repo programs are run with their current working
+   directory set to a temporary directory to which they can freely write and
+   which will be automatically removed when they have completed.
  * `<secrets-path>` is the file containing the GitHub secret which guarantees
    that hooks are coming from your GitHub repository and not a malfeasant. Note
    that leading and trailing whitespace in this file is ignored.
--- a/snare
+++ b/snare
@@ -27,10 +27,11 @@ class HookHandler(BaseHTTPRequestHandler):
 
         path = os.path.join(repo_scripts_dir, owner, repo)
         if os.path.exists(path):
-            with tempfile.NamedTemporaryFile() as jsonf:
-                jsonf.write(data_str.encode())
-                event = self.headers["X-GitHub-Event"]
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -19,7 +19,9 @@ where:
  * `<repo-programs-dir>` is the directory where the per-repo programs are
    stored. For a repository `repo` owned by `user` the command
    `<repo-programs-dir>/<user>/<repo> <event> <path to GitHub JSON>` will be
-   run.
+   run. Note that per-repo programs are run with their current working
+   directory set to a temporary directory to which they can freely write and
+   which will be automatically removed when they have completed.
```
vext01: A nitpick. I think we can make this more easily understood with:
```
Note that per-repo programs are run in a temporary directory which will be automatically removed when they have completed.
```

I don't think it's necessary to say that the directory is writable. The user will assume this.

vext01:
Aside from one that one small comment, LGTM.

README.md
ltratt: If we don't say it's writeable, they may assume we've done some sort of `chroot` tricks and things are readable only. It certainly doesn't hurt to make it clear that users can write to this directory as they wish.

vext01: I'm not sure what chroot has to do with permissions. You can have a writable chroot.

If you want to say it's writeable, how about:
> Note that per-repo programs are run in a (writeable) temporary directory which will be automatically removed when they have completed.

ltratt: I think we might be in danger of crossing into nit-picking territory now :) I don't think there's anything inaccurate, missing, or misleading in the text.

vext01: Well, I did open with "A nitpick" :)

I think it's a bit wordy personally. I had to re-read it a couple of times to absorb it.

Compare my suggestion:
> Note that per-repo programs are run in a (writeable) temporary directory which will be automatically removed when they have completed.

With the original:
> Note that per-repo programs are run with their current working directory set to a temporary directory to which they can freely write and which will be automatically removed when they have completed.

The latter is quite jarring.

I don't feel strongly enough to argue further, mind.

ltratt: Amusingly I started with a sentence that said "run in a directory" before adding the stuff about CWD and writeable and so on. Let's keep it as it is and revise in the future.

vext01: I disagree with the decision, but so be it :P

PR #6, opened 2019-11-21
Fix markdown.

ltratt:


vext01:
Glad I'm not the only one who can't get links right in md!

PR #7, opened 2020-01-22
Rewrite snare in Rust.

ltratt:
This does everything the Python version can and a bit more (notably it can send emails when a per-repo program fails containing its stderr/stdout). However, the intention is to become quite a bit more sophisticated in the future, so the architecture of this is more complicated than it really needs to be at the moment.

Architecturally this system is split in two:
  * The `httpserver` listens for incoming hooks, checks that they're valid, and
    adds them to a `Queue`.
  * The `jobrunner` pops elements from the `Queue` and runs them in parallel.

These two components run as two different threads: the `httpserver` writes a solitary byte to an "event pipe" to wake up the `jobrunner` when the queue has new elements. This allows the `jobrunner` to use a single interface for listen for completed jobs as well as new jobs.

Since this is intended to be a fairly traditional Unix daemon (albeit written in Rust), most of the system goes out of its way to keep running even in the face of annoying errors. There are still a few things that can cause it to panic, but those can be reduced/removed in the future.


vext01:
Before I review the code: why the re-write? I'd have thought that for a task like this, a Python script would be much easier to maintain and deploy.

ltratt:
IMHO it's very difficult, indeed borderline impossible, to make a truly reliable daemon in exception-based languages because you can't really know which lines of code might explode at run time.

This rewrite aims to move snare in the direction of being as close to bullet proof as is practical, which as you'll see means carefully handling a lot of not-very-likely error conditions. It's not all the way there yet, by any means, but hopefully this moves things in a plausible direction.

vext01:
OK, I buy that :)

I'll proceed with review after lunch.

README.md
```diff
@@ -10,11 +10,16 @@ can then perform whatever actions it wants.
 `snare` requires three command-line arguments to be specified:
 
 ```
-Usage: snare -p <port> -r <repo-programs-dir> -s <secrets-path>
+Usage: snare [-e email] [-j <maxjobs>] -p <port> -r <reposdir> -s <secretspath>
```
vext01: `<max-jobs>` might be more aesthetically pleasing? Similarly with other double-barrelled argument names.

Cargo.toml
```diff
@@ -0,0 +1,40 @@
...
+name = "snare"
+version = "0.1.0"
+authors = ["Laurence Tratt <laurie@tratt.net>"]
+edition = "2018"
+
+[[bin]]
+doc = false
+name = "snare"
+
+[lib]
```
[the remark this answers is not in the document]
ltratt: I think that's just the easiest way to structure things internally. It certainly won't go on crates.io!

README.md
ltratt: Fixed in https://github.com/softdevteam/snare/pull/7/commits/d468f70cf288a13b8556b296a41c3845cbd195e5.

README.md
```diff
@@ -10,11 +10,16 @@ can then perform whatever actions it wants.
...
 ```
 
 where:
 
+ * `<email>` is an (optional) email address to which any errors running per-repo
+   programs will be sent (warning: full stderr/stdout will be sent, so consider
```
vext01: Had to double take this sentence, but I can't see a rewording which makes it any easier.

src/lib/config.rs
```diff
@@ -0,0 +1,119 @@
...
+
+use getopts::Options;
+
+pub struct Config {
+    /// The maximum number of parallel jobs to run.
+    pub maxjobs: usize,
+    /// The port to listen on.
+    pub port: u16,
+    /// A *fully canonicalised* path to the directory containing per-repo programs.
+    pub reposdir: String,
```
vext01: Is that better as a `PathBuf`?

src/lib/config.rs
```diff
@@ -0,0 +1,119 @@
...
+    /// The GitHub secret used to validate requests.
+    pub secret: String,
```
vext01: I'm not suggesting you do this if you don't want, but I like to use `secstr` for in-memory secrets:
https://crates.io/crates/secstr

src/lib/config.rs
```diff
@@ -0,0 +1,119 @@
...
+    /// An optional email address to send errors to.
+    pub email: Option<String>,
```
vext01: Should this be a vector to allow multiple recipients?

src/lib/config.rs
```diff
@@ -0,0 +1,119 @@
...
+}
+
+impl Config {
+    /// Create a `Config`.
```
vext01: I'd probably drop that comment. It's obvious even by my standards :)

Cargo.toml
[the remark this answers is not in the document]
vext01: It seems a little odd. Why not use a `main.rs` with `mod` lines? Then you can drop `[lib]` and `[bin]`.

src/lib/httpserver.rs
```diff
@@ -0,0 +1,143 @@
...
+use hyper::service::{make_service_fn, service_fn};
+use hyper::{self, body::Bytes, Body, Request, Response, Server, StatusCode};
+use json;
+use percent_encoding::percent_decode;
+use sha1::Sha1;
+
+use crate::{queue::QueueJob, Snare};
+
+pub(crate) async fn serve(snare: Arc<Snare>) {
+    let addr = SocketAddr::from(([0, 0, 0, 0], snare.config.port));
```
vext01: Nit for another time: allow customisation of bind address.

src/lib/httpserver.rs
```diff
@@ -0,0 +1,143 @@
...
+/// leads to a blanket `Err(())` so that an attacker cannot deduce anything about the cause of the
+/// failure to authenticate.
+async fn authenticate(req: Request<Body>, snare: &Arc<Snare>) -> Result<Bytes, ()> {
+    // Extract the string 'def' from "X-Hub-Signature: abc=def"
+    let sig = req
+        .headers()
+        .get("X-Hub-Signature")
+        .ok_or(())?
+        .to_str()
+        .map_err(|_| ())?
```
vext01: Since performance isn't key for snare, why don't we make the error type `Box<Error>`, then you can kill all of the `map_err`s.

src/lib/httpserver.rs
```diff
@@ -0,0 +1,143 @@
...
+    // Verify that this request was signed by the same secret that we have.
+    let mut mac = Hmac::<Sha1>::new_varkey(snare.config.secret.as_bytes()).map_err(|_| ())?;
+    mac.input(&*data);
+    mac.verify(&hex::decode(sig).map_err(|_| ())?)
+        .map_err(|_| ())?;
+
+    Ok(data)
+}
+
+/// Parse `pl` into JSON, and return `(<JSON as a String>, <repo owner>, <repo name>)`.
```
vext01: If I were reading documentation, I'd be left guessing that `pl` stands for `payload`. For clarity, I'd just use the full word as the argument name.

src/lib/httpserver.rs
```diff
@@ -0,0 +1,143 @@
...
+fn parse(pl: Bytes) -> Result<(String, String, String), ()> {
+    // The body sent by GitHub starts "payload=" before then containing JSON encoded using the URL
+    // percent format.
+
+    // First check that the string really does begin "payload=".
```
vext01: Would it be easier to use `.starts_with()` here and below?

src/lib/jobrunner.rs
```diff
@@ -0,0 +1,409 @@
...
+    fcntl::{fcntl, FcntlArg, OFlag},
+    poll::{poll, PollFd, PollFlags},
+};
+use tempfile::{tempdir, tempfile, NamedTempFile, TempDir};
+use whoami::{hostname, username};
+
+use crate::{queue::QueueJob, Snare};
+
+const READBUF: usize = 8 * 1024; // Bytes. Should be >= PIPE_BUF for performance reasons.
+const WAIT_TIMEOUT: i32 = 1; // Seconds.
```
vext01: For these timeouts it might be possible to use `Duration::from_secs()` (which is const), then you don't need a comment explaining the unit.

src/lib/jobrunner.rs
```diff
@@ -0,0 +1,409 @@
...
+    /// The running jobs, with `0..2 *num_running + 1` entries. Each pair of entries are (stderr,
+    /// stdout) for the corresponding `running` `Job` (i.e. `running[0]` has its stderr entry at
+    /// `pollfds[0]` and stdout entry at `pollfds[1]`). The `+ 1` entry is the event file
+    /// descriptor that allows the HTTP server thread to wake up the JobRunner thread.
+    pollfds: Vec<PollFd>,
+}
+
+impl JobRunner {
+    fn new(snare: Arc<Snare>) -> Result<Self, Box<dyn Error>> {
+        if snare.config.maxjobs == std::usize::MAX {
```
vext01: Wouldn't it be better to check this at argument parsing time?

src/lib/jobrunner.rs
```diff
@@ -0,0 +1,409 @@
...
+            eprintln!(
+                "The maximum number of possible jobs is {}",
+                std::usize::MAX - 1
+            );
+            process::exit(1);
```
vext01: I saw a function for exiting with an error earlier. Can we use it here?

src/lib/jobrunner.rs
```diff
@@ -0,0 +1,409 @@
...
+        });
+        Ok(JobRunner {
+            snare,
+            running,
+            num_running: 0,
+            pollfds,
+        })
+    }
+
+    fn attend(&mut self) {
```
vext01: This function probably does need a doc string. What are we attending?

src/lib/jobrunner.rs
```diff
@@ -0,0 +1,409 @@
...
+            } else {
+                DEFAULT_TIMEOUT
+            };
+            poll(&mut self.pollfds, timeout * 1000).ok();
+
+            // See if any of our active jobs have events. Knowing when a pipe is actually closed is
+            // surprisingly hard. https://www.greenend.org.uk/rjk/tech/poll.html has an interesting
+            // suggestion which we adapt slightly here.
+            for i in 0..self.snare.config.maxjobs {
+                // stderr
```
vext01: partial comment left in?

vext01: Maybe not though, because there's another similar one below for stdout.

So what about stderr?

Nit: other comments in the file have a full stop.

src/lib/jobrunner.rs
```diff
@@ -0,0 +1,409 @@
...
+                        return Err(Some(qj));
+                    }
+                    p
+                }
+                Err(_) => return Err(Some(qj)),
+            },
+            Err(_) => return Err(Some(qj)),
+        };
+
+        // We combine the child processe's stderr/stdout and write them to an unnamed temporary
```
vext01: Is that a typo `processe's`?  Not sure myself.

Plural of process is "processes", so I'd guess *processes'*?

src/lib/jobrunner.rs
```diff
@@ -0,0 +1,409 @@
...
+                    }
+                }
+            }
+        }
+
+        Err(Some(qj))
+    }
+
+    /// After a job has been inserted / removed from `self.running`, this function must be called
+    /// so that `poll()` is called with upto date file descriptors.
```
vext01: `upto` -> `up to`

src/lib/jobrunner.rs
```diff
@@ -0,0 +1,409 @@
...
+
+    /// If the user has specified an email address, send the contents of
+    fn sendemail(&self, mut file: &File) {
+        if let Some(ref toaddr) = self.snare.config.email {
+            let mut buf = Vec::new();
+            buf.extend("Subject: snare error\n\n".as_bytes());
+            file.seek(SeekFrom::Start(0)).ok();
+            file.read_to_end(&mut buf).ok();
+
+            let fromea = match EmailAddress::new(format!("{}@{}", username(), hostname())) {
```
vext01: `fromea` -> `from_ea` perhaps?

src/lib/jobrunner.rs
```diff
@@ -0,0 +1,409 @@
...
+    /// Log a warning to the user. This might be to stderr, a log file, or email.
+    fn warning(&self, msg: &str) {
+        eprintln!("Warning: {}\n", msg);
+    }
+}
+
+struct Job {
+    child: Child,
+    /// This TempDir will be dropped, and its file system contents removed, when this Job is dropped.
+    _tempdir: TempDir,
```
vext01: So is `_tempdir` needed? The underscore suggests it isn't used?

src/lib/jobrunner.rs
```diff
@@ -0,0 +1,409 @@
...
+    fcntl(
+        fd,
+        FcntlArg::F_SETFL(unsafe { OFlag::from_bits_unchecked(flags) }),
+    )?;
+    Ok(())
+}
+
+pub(crate) fn attend(snare: Arc<Snare>) -> Result<(), Box<dyn Error>> {
+    let mut rn = JobRunner::new(snare)?;
+    thread::spawn(move || rn.attend());
```
vext01: Do we get an appropriate error message if this spawn fails? I guess you have to check something in the join handle?

vext01:
That's my somewhat superficial review done.

I can't comment on the tokio code. (Last time I used that, it was using futures, but now seems to be using the new async stuff.)

You might consider adding module-level `//!` comments to some/all files? Only if you think it's necessary mind.

src/lib/config.rs
ltratt: Not really.

src/lib/config.rs
ltratt: OK, might make sense for a future PR.

src/lib/config.rs
ltratt: This is a temporary feature which will be expanded in the future, so it's OK for now I think.

src/lib/httpserver.rs
ltratt: It's stored as bytes, so it's not quite that easy :(

src/lib/jobrunner.rs
ltratt: I disagree: we'd still need a unit in the comment whatever the case, otherwise the `const` definition doesn't stand on its own.

src/lib/jobrunner.rs
ltratt: Good point. Fixed in https://github.com/softdevteam/snare/pull/7/commits/f848da2f90a44631a17e6acb5d3ababcfd874edf.

src/lib/jobrunner.rs
ltratt: Fixed in https://github.com/softdevteam/snare/pull/7/commits/5ed09d85cc26aa0d3de30c49da1695f48513b4e0.

src/lib/jobrunner.rs
ltratt: Fixed in https://github.com/softdevteam/snare/pull/7/commits/6283b7a29a721248f11d0b430033df2bb7cb4eec.

src/lib/jobrunner.rs
ltratt: It turns out it should be "up-to-date". I didn't know that! Fixed in https://github.com/softdevteam/snare/pull/7/commits/5a1af6c45f75b99330e7dd20c9418ca26cadacfc.

src/lib/jobrunner.rs
ltratt: It's so that it's kept alive until `Job` is dropped.

src/lib/httpserver.rs
vext01: Right, but you are about to convert it to a string anyway, so why not do that, then check the prefix?

src/lib/jobrunner.rs
vext01: I don't understand. `Duration::from_secs()` tells the reader the unit is seconds.

ltratt: You have to look for where it's used to realise that. By adding it in the comment we make it clearer.

src/lib/httpserver.rs
ltratt: It's a bit complex but we don't convert that to a string directly, so it makes sense to efficiently strip just the bit off we need. [If we did convert to a string, we'd do a UTF-8 check, which we'd then have to do again after `percent_encode` has done its thing. That seems a bit wasteful.]

src/lib/jobrunner.rs
vext01: So now we have another variation: `process's`. I don't know which is right, but to be clear that's another variation to what was suggested above.

It's my understanding that to make a word that already ends with an `s` posessive, you add a `'`.

But as you know, I'm not good at grammar!

ltratt: I refer you to http://www.apostrophe.org.uk/page2.html -- which is what I always have to check as I can never remember either!

src/lib/httpserver.rs
vext01: fair enough.

src/lib/jobrunner.rs
vext01: Eh? I sense crossed wires.

```
const WAIT_TIMEOUT: Duration = Duration::seconds(1);
```

src/lib/jobrunner.rs
vext01: Ah! Cool!

But wait!

> ... however, if there are two or more dogs, companies or Joneses in our example, the apostrophe comes after the 's':
>
>    the dogs' bones
>    the companies' logos
>    Joneses' bakeries

 * Singular *process*
 * Plural *processes*
 * Possessive plural *processes'*

(I know, this is getting absurd :P )

vext01: Wait, I see what's going on here.

We are combining stdout and stderr for one process, not the stdout and stderr for many processes. So you are right!

src/lib/jobrunner.rs
ltratt: Mea culpa! Yes, I like the idea, so I tried it. Unfortunately `Duration::as_millis()` returns a `u128` when I need an `i32`, so this accidentally made things worse :( Oh well. I've still beefed up the comments in https://github.com/softdevteam/snare/pull/7/commits/e636f4a5b4ff4a18e6085a845fe1a560b1c54744 to get something from the exercise.

vext01: Ah well. It was worth a shot!

ltratt:
I think those are all of the addressable comments. If you're OK, it might be time to squash this monster down...

src/lib/jobrunner.rs
vext01: Don't you check for error when `join()`ing via the handle?

ltratt: There is no `join` here, as that would block the HTTP thread!

vext01:
In addition to the thread joining comment I just added, I think you've left me hanging on a few other comments.

 * The "lib-ness" of the program and directory structure.
 * Kill that obvious comment?
 * Kill uses of `map_err()` with `Box<Error>`.
 * Use of your `error()` function in the jobrunner.

Cargo.toml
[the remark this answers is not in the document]
ltratt: I thought I tried that and it didn't work? Am I missing something?

src/lib/config.rs
ltratt: Agreed, but since it's a public(ish) function, it does feel better than nothing.

src/lib/httpserver.rs
ltratt: It's partly to emphasise that "this is not an error you can recover from because you can't even tell what it was".

src/lib/jobrunner.rs
ltratt: Not really, that other function isn't really exportable.

src/lib/jobrunner.rs
ltratt: There are two parts to that `for` loop: the bit that handles stderr and the bit that handles stdout. They're easy to confuse, hence the minor comment.

src/lib/jobrunner.rs
ltratt: Potato, tomato, potato, tomato :p

src/lib/jobrunner.rs
vext01: Oh. Hrm.

So supposing that thread panics *or* fails to spawn (i think the cases are indistinguishable), would another part of the program notice and print an error?

If not, that's bad. Perhaps we need to store the handle somewhere and check for error at such a time when it is safe (in that it won't block anything).

ltratt: As i said, I don't know how to do this in the current API...

vext01:
Not sure if this bothers you:
```
error: `core::slice::<impl [T]>::len` is not yet stable as a const fn
   --> /home/vext01/.cargo/registry/src/github.com-1ecc6299db9ec823/bytes-0.5.3/src/bytes.rs:121:18
    |
121 |             len: bytes.len(),
    |                  ^^^^^^^^^^^
    |
    = help: add `#![feature(const_slice_len)]` to the crate attributes to enable
```

vext01:
Never mind, `rustup update` fixed that.

ltratt:
I think I've gone through the unaddressed comments now.

src/lib/jobrunner.rs
vext01: Indeed. This is annoying.

I tried making `attend()` panic to see what happens:
```
     Running `target/debug/snare -p 3333 -r /tmp -s /tmp/s`
thread '<unnamed>' panicked at 'fish', src/lib/jobrunner.rs:65:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

And then the program hangs.

Oh well.

ltratt: I think we have to accept this as a limitation of the API, at least for now. Maybe we'll work out how to do it in the future.

Cargo.toml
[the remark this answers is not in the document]
vext01: Maybe. This setup would be more standard I think:

```
$ find src
src
src/main.rs
src/config.rs
src/httpserver.rs
src/jobrunner.rs
src/queue.rs

$ cat src/main.rs
mod config;
mod httpserver;
mod jobrunner;
mod queue;

fn main() {}

$ cat Cargo.toml 
[package]
name = "bintest"
version = "0.1.0"
edition = "2018"

$ cargo build
   Compiling bintest v0.1.0 (/tmp/bintest)
   Finished dev [unoptimized + debuginfo] target(s) in 0.35s
```

Unless I missed something you shouldn't need `[lib]` or `[bin]` in `Cargo.toml`.

src/lib/httpserver.rs
vext01: OK, fair enough.

Cargo.toml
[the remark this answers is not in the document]
ltratt: I'm an idiot! Fixed in https://github.com/softdevteam/snare/pull/7/commits/74ed3517892ea7ee7da5efd9fbb6db0910a00b53.

src/lib/jobrunner.rs
vext01: OK, I won't obsess on this one.

vext01:
OK, enough bikeshedding on my part.

If you are ready, please squash.

vext01:
bors ping

ltratt:
bors try

ltratt:
Please check https://github.com/softdevteam/snare/pull/7/commits/5df76bffc6bfa8f5914d903c95a019a7453502ce -- if you're OK with that, I'll squash and then we can try bors for real!

vext01:
Did you mean to only `check` and not `build`?

ltratt:
I couldn't see any advantage of `build` over `check` when we're not running anything, but I'm happy to be shown to be wrong!

vext01:
Ah i see.

Well it's up to you. I've seen cases where you get errors at build time, but not at check time. Usually due to obscure linkage usage, which I doubt we will use in snare TBH.

I'll roll with whatever you decide is best ;)

ltratt:
> I've seen cases where you get errors at build time, but not at check time. Usually due to obscure linkage usage, which I doubt we will use in snare TBH.

I can buy that. Fixed in https://github.com/softdevteam/snare/pull/7/commits/938baacf6b87ced6b147f19abae293a0134e3f00.

vext01:
LGTM. Please squash.

ltratt:
Squashed.

PR #9, opened 2020-01-23
Split apart binding to a port from running the HTTP server.

ltratt:
This paves the way for us to, in the future, change the UID of snare from root to a non-root user.


vext01:
Can't we just bind to a port > 1024? Then we wouldn't need to be root.

ltratt:
> Can't we just bind to a port > 1024? Then we wouldn't need to be root.

People might want to bind to such a port -- I think we should support that (at some point).

vext01:
> People might want to bind to such a port

Personally, I'd run the program unprivileged and use a firewall rule.

Either way, code looks good. Shall I merge?

ltratt:
Yes please.

PR #10, opened 2020-01-24
Introduce a simple function to allow the program to exit fatally.

ltratt:


vext01:
One thing that comes to mind: what if there's no language-level error per-se, but we want to exit? I'd probably have made your function accept a `String` and then the callsite can format it, but it's no big deal.

If that situation arises, we can deal with it then.

bors r+

ltratt:
> One thing that comes to mind: what if there's no language-level error per-se, but we want to exit? I'd probably have made your function accept a String and then the callsite can format it, but it's no big deal.
>
>If that situation arises, we can deal with it then.

Agreed.

PR #11, opened 2020-01-24
Remove the pointlessly paranoid "always check the queue every minute".

ltratt:
This paranoia was entirely pointless, because if it was ever necessary, we would simply get stuck elsewhere. The idea was that, perhaps, sometimes we might not be notified that there's a job on the queue so we should (infrequently) check the queue even if we haven't been told there's an update. However, this could only happen if poll() failed. If that happened, then our ability to read stderr/stdout from child processes would also fail, at which point we'd be stuffed. But poll() isn't very likely to fail, so the paranoia is completely
pointless: we might  as well simplify the code and be done with it!

### DIFF
--- a/src/jobrunner.rs
+++ b/src/jobrunner.rs
@@ -21,14 +21,9 @@ use crate::{queue::QueueJob, Snare};
 
 /// The size of the temporary read buffer in bytes. Should be >= PIPE_BUF for performance reasons.
 const READBUF: usize = 8 * 1024;
-/// Maximum time to wait in `poll` (in seconds) while waiting for child processes to terminate.
+/// Maximum time to wait in `poll` (in seconds) while waiting for child processes to terminate
+/// and/or because there are jobs on the queue that we haven't been able to run yet.
 const WAIT_TIMEOUT: i32 = 1;
-/// Maximum time to wait in `poll` (in seconds) while there are jobs on the queue that we haven't
-/// been able to run.
-const RECHECK_TIMEOUT: i32 = 5;
-/// Maximum time to wait in `poll` (in seconds). This is really an exercise in "just in case"
-/// paranoia in case we somehow weren't notified of data to read or similar.
-const DEFAULT_TIMEOUT: i32 = 60;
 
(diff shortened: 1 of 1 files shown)


vext01:
Maybe I don't understand, but if you can't miss a notification, why don't you pass `INFTIM` as the timeout?

ltratt:
> Maybe I don't understand, but if you can't miss a notification, why don't you pass INFTIM as the timeout?

Sometimes we're also waiting on processes to finally terminate. That's not something that `poll` can tell us about, so we have to wake up every so often and explicitly check. [`kqueue` can check for this, but it's not portable :(]

vext01:
> Sometimes we're also waiting on processes to finally terminate.

I see. Would the signal mask argument of `ppoll()` help (to catch `SIGCHILD`)?

vext01:
If I understand correctly, a signal-based approach would be better, as at the moment you will only check for dead processes once every 16 minutes.

ltratt:
> If I understand correctly, a signal-based approach would be better, as at the moment you will only check for dead processes once every 16 minutes.

No, we check every second.

I'm not sure wheterh `ppoll` and a mask works. I need to look into it, but I think that would be for a future PR.

vext01:
> No, we check every second.

I must be missing something. You wait up to either `TIMEOUT * 1000` or `RECHECK_TIMEOUT * 1000` where the constants are 1 and 5 seconds respectively (according to the comment next to them).

vext01:
I see the confusion. poll takes milliseconds.

So you wait either 1 second, 5 seconds or `INFTIM` (-1), depending on some stuff.

I don't really understand the code, so if that's intentional, let's move on?

ltratt:
Now that I think about it, I think I can simplify this even further. Bear with me...

ltratt:
Take a gander at https://github.com/softdevteam/snare/pull/11/commits/205df41e23b3f78353d35620dbe4155576f01436

vext01:
LGTM, please squash.

ltratt:
Squashed.

PR #12, opened 2020-01-27
Move to using a configuration file.

ltratt:
At the moment, the amount of command-line switches we have is just about bearable, but this is likely to change profoundly in the future. This commit thus moves to a model of a simple configuration file.

@vext01 A question for you. My plan is for the next PR to move to a config format where you can start specifying this:

```
github {
  match "*" {
    timeout = 60
    secret = "..."
  }

  match "user/repor" {
    timeout = 600
  }
}
```

i.e. something like PF where user can specify different settings per-repo, and that happens by matching the config file from top to bottom with "last matching setting wins".

Given the curly brackets (which seem the best way of specifying this config), would it be more idiomatic to have statements end with a semi-colon `;`? I can't quite make up my mind.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,13 +4,21 @@ version = "0.1.0"
 authors = ["Laurence Tratt <laurie@tratt.net>"]
 edition = "2018"
 
+[build-dependencies]
+cfgrammar = "0.5"
+lrlex = "0.5"
+lrpar = "0.5"
+
 [dependencies]
+cfgrammar = "0.5"
 getopts = "0.2"
 hex = "0.4"
 hmac = "0.7"
 hyper = "0.13"
 json = "0.12"
 lettre = "0.9"
+lrlex = "0.5"
+lrpar = "0.5"
 mio = "0.6"
 nix = "0.16"
 num_cpus = "1.12"
@@ -19,6 +27,7 @@ secstr = "0.3"
 sha-1 = "0.8"
 tempfile = "3.1"
 tokio = { version = "0.2", features = ["full"] }
+users = "0.9"
 whoami = "0.7"
 
 [profile.release]
--- a/README.md
+++ b/README.md
@@ -7,32 +7,55 @@ can then perform whatever actions it wants.
 
 ## Basic usage
 
-`snare` requires three command-line arguments to be specified:
+`snare` has the following command-line format:
 
 ```
-Usage: snare [-e email] [-j <max-jobs>] -p <port> -r <repos-dir> -s <secrets-path>
+Usage: snare [-c <config-path>]
 ```
 
 where:
 
- * `<email>` is an (optional) email address to which any errors running per-repo
-   programs will be sent (warning: full stderr/stdout will be sent, so consider
-   carefully whether these have sensitive information or not).
- * `<max-jobs>` is an (optional) non-zero integer specifying the maximum number
-   of jobs to run in parallel.
- * `<port>` is a port number (e.g. 4567).
- * `<repo-programs-dir>` is the directory where the per-repo programs are
-   stored. For a repository `repo` owned by `user` the command
-   `<repo-programs-dir>/<user>/<repo> <event> <path to GitHub JSON>` will be
-   run. Note that per-repo programs are run with their current working
-   directory set to a temporary directory to which they can freely write and
-   which will be automatically removed when they have completed.
- * `<secrets-path>` is the file containing the GitHub secret which guarantees
-   that hooks are coming from your GitHub repository and not a malfeasant. Note
-   that leading and trailing whitespace in this file is ignored.
+ * `<config-path>` is a path to a `snare.conf` configuration file. If not
+   specified explicitly, the following locations will be searched, in order:
+     * `/etc/snare.conf/`
+     * `~/.snare.conf`
 
-The most important part of this are the per-repo programs. For example,
-`snare`'s GitHub repository is
+
+## Configuration file
+
+The configuration file supports the following options:
+
+ * `email = "<address>"` optionally specifies an email address to which any
+   errors running per-repo programs will be sent (warning: full stderr/stdout
+   will be sent, so consider carefully whether these have sensitive information
+   or not).
+ * `maxjobs = <int>` is an (optional) non-zero integer specifying the maximum
+   number of jobs to run in parallel. Defaults to the number of CPUs in the
+   machine.
+ * `port = <int>` is a mandatory port number to listen on (e.g. 4567).
+ * `reposdir = "<path>"` is the directory where the per-repo programs are
+   stored. For a repository `repo` owned by `user` the command
+   `<reposdir>/<user>/<repo> <event> <path to GitHub JSON>` will be run. Note
+   that per-repo programs are run with their current working directory set to a
+   temporary directory to which they can freely write and which will be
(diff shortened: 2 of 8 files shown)


vext01:
Wouldn't using a toml configuration file be more rusty? Then you don't have to worry about parsing at all.

ltratt:
This isn't supposed to be Rusty -- it's an old fashioned Unix daemon that just so happens to be written in Rust :) The plan is also for the configuration file to be extended in such a way that toml would be completely inappropriate.

vext01:
OK. I'd just use a standard format myself, but if you want to go through the pain of writing a parser, knock yourself out :P

vext01:
re semi-colons: no preference.

ltratt:
The parser isn't too painful, and you get nice error messages, so I'm fine with it :)

vext01:
Fair enough. You are braver than I!

ltratt:
Anything stopping this being bors'd?

Cargo.toml
```diff
@@ -4,13 +4,21 @@ version = "0.1.0"
 authors = ["Laurence Tratt <laurie@tratt.net>"]
 edition = "2018"
 
+[build-dependencies]
+cfgrammar = "0.5"
+lrlex = "0.5"
+lrpar = "0.5"
+
 [dependencies]
+cfgrammar = "0.5"
```
vext01: Is it correct that `cfgrammar` is both a build and a run dependency?

README.md
```diff
@@ -7,31 +7,53 @@ can then perform whatever actions it wants.
 
 ## Basic usage
 
-`snare` requires three command-line arguments to be specified:
+`snare` has the following command-line format:
 
 ```
-Usage: snare [-e email] [-j <max-jobs>] -p <port> -r <repos-dir> -s <secrets-path>
+Usage: snare [-c <config path>]
```
vext01: `<config-path>` with a hyphen, like before?

README.md
```diff
@@ -7,31 +7,53 @@ can then perform whatever actions it wants.
...
+```
+port = <port>
+reposdir = "<path>"
+secret = "<secret>"
+```
+
+
+## Per-repo programs
+
+The most important part of snare are the per-repo programs. For example,
```
vext01: > the most important part of snare are

Should it be:

> the most important part of a snare deployment are

As the programs are not part of snare, per-se, right? The user provides those.

build.rs
```diff
@@ -0,0 +1,14 @@
+use cfgrammar::yacc::YaccKind;
+use lrlex::LexerBuilder;
+use lrpar::CTParserBuilder;
+
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let lex_rule_ids_map = CTParserBuilder::<u8>::new_with_storaget()
```
vext01: I know this is upstream code, but the term `storaget` is unfortunate. I guess it means "storage T", but it's hard not to read it as one word.

vext01: I thought the same about `reposdir` fwiw. Took a second glance to understand the meaning. An underscore would help, but I'm not going to insist ;)

src/config.rs
```diff
@@ -28,87 +42,154 @@ impl Config {
...
+        if secret.is_none() {
+            fatal("A secret must be specified");
+        }
 
         Config {
-            maxjobs,
-            port,
-            reposdir,
-            secret,
+            maxjobs: maxjobs.unwrap(),
```
vext01: If you unwrap your config directives above, you can drop this hunk. E.g.:

```
let port = port.unwrap_or_else(|| fatal(...));
```

src/config.rs
```diff
@@ -126,3 +207,33 @@ fn usage(prog: &str) -> ! {
     .ok();
     process::exit(1)
 }
+
+/// Try to find a `snare.conf` file.
+fn search_snare_conf() -> Option<PathBuf> {
+    for cnd in SNARE_CONF_SEARCH {
+        if cnd.starts_with("~/") {
+            let mut homedir =
+                match get_user_by_uid(get_current_uid()).map(|x| x.home_dir().to_path_buf()) {
```
vext01: You could use `$HOME` from the environment, and drop the dependency on `users`?

src/main.rs
```diff
@@ -33,19 +33,25 @@ pub(crate) struct Snare {
     event_write_fd: RawFd,
 }
 
-/// Exit with a fatal error, printing the contents of `err`.
-fn fatal<E: Into<Box<dyn Error>> + Display>(msg: &str, err: E) -> ! {
-    eprintln!("{}: {}", msg, err);
+/// Exit with a fatal error.
+fn fatal(msg: &str) -> ! {
+    debug_assert!(msg.chars().last() != Some('.'));
```
vext01: Is this to ensure that error messages all have the same format, without a full-stop?

vext01:
A handful of very minor comments.

I can't vouch for the grammar.

Sorry for the delay, I thought you were just asking about the semi-colon, and didn't realise there was actually code to review! :P

Cargo.toml
ltratt: Yes.

README.md
ltratt: Fixed in https://github.com/softdevteam/snare/pull/12/commits/b32b12216bad9fbea17e2d48169eafe7f3a9475f.

README.md
ltratt: Fixed in https://github.com/softdevteam/snare/pull/12/commits/0cf50f27281b617a03b5b5f97503de499429e5f9.

build.rs
ltratt: `StorageT` is what it is. `reposdir` is more a matter of taste. We disagree on that ;)

src/config.rs
ltratt: True. I'm not sure it's an improvement or not. Maybe I might change this in a future PR when I've slept on it!

src/config.rs
ltratt: I copied this idea from extsmaild -- I no longer remember why I did this other than I think I had some reason to do so. Maybe because things like `sudo` don't get `$HOME` correct?

src/main.rs
ltratt: Correct.

build.rs
vext01: `StorageT` is fine (and not at question), but `storaget` is hard to read.

All I'm saying is, sometimes an underscore would help to delimit words.

E.g. `storage_t`.

(I'm not going to insist)

src/config.rs
vext01: It's minor really. I don't care too much.

build.rs
ltratt: You're asking for a change in another repo, so the answer for this PR is definitely "out of scope" :p

src/config.rs
vext01: Hrm, yeah not sure what the desired outcome is, but `sudo echo $HOME` does not print `/root`.

Again I don't feel strongly.

ltratt: OK, then I think `get_user_by_uid` is the right thing to do, since it's more reliable than `$HOME`.

ltratt:
Squashed.

build.rs
vext01: Yes, I wasn't asking for a change, only commenting.

> I know this is upstream code, but ...

PR #13, opened 2020-01-28
Allow per-repository configuration settings

ltratt:
This change allows users to specify most settings on a per-repository basis. The README contains more details, but the most important part can be seen in this example configuration file:

```
  port = <port>
  github {
    reposdir = "<path>"
    match ".*" {
      email = "abc@def.com"
    }
    match "a/b" {
      email = "ghi@jkl.com"
    }
  }
```

`match` blocks contain a regex matching against a repository name, and are evaluated in order from top to bottom with each successful match overriding previous settings. This allows users to specify defaults which are only overridden for specific repositories. In this example the repository `c/d` will have errors sent to `abc@def.com` but `a/b` will have errors sent to `ghi@jkl.com`.

This was a more involved change than I had hoped, in part because it changed the authentication mechanism. In short, in order to authenticate a webhook request, we have to load the data, find the relevant config (since `secret` can now be specified per-repository), and then authenticate.


ltratt:
I'm sure you'll find other things as well, but those last couple of commits will need squashing.

vext01:
One thing before I review the code:

> overriding previous settings

I'm probably over-thinking this, but I can imagine a scenario where some boss man wants to receive all emails, and then his minions who work on individual projects, only want to receive mail about their project.

It might be tempting to have a:
```
match ".*" {
      email = "bigboss@bigbucks.com"
    }
```
But the individual project owners would override his email.

One way you could cater for this would be to make email a list, and allow variables (like in `vm.conf`.

Then you could have something like:
```
$bossman = "bigboss@bigbucks.com"

github {
  match "a/b" {
      email = [$bossman, "jim@bigbucks.com"]
    }
}
```

Or an easier, but less flexible, approach would be to have a global `extra_emails` config option.

I'm not suggesting we action this now, but food for thought.

README.md
```diff
@@ -23,33 +23,75 @@ where:
...
 
- * `email = "<address>"` optionally specifies an email address to which any
-   errors running per-repo programs will be sent (warning: full stderr/stdout
-   will be sent, so consider carefully whether these have sensitive information
-   or not).
  * `maxjobs = <int>` is an (optional) non-zero integer specifying the maximum
    number of jobs to run in parallel. Defaults to the number of CPUs in the
    machine.
  * `port = <int>` is a mandatory port number to listen on (e.g. 4567).
+ * `github { ... }` specifies GitHub specific options.
```
vext01: This is to allow for other (non-github) webhook senders to be integrated into snare?

ltratt:
I do agree that we will one day want to allow multiple email addresses but not now. [I'm not sure whether variables are useful in something as simple as snare, to be honest.]

README.md
ltratt: Correct.

README.md
```diff
@@ -23,33 +23,75 @@ where:
...
    stored. For a repository `repo` owned by `user` the command
    `<reposdir>/<user>/<repo> <event> <path to GitHub JSON>` will be run. Note
    that per-repo programs are run with their current working directory set to a
    temporary directory to which they can freely write and which will be
    automatically removed when they have completed.
- * `secret = "<secret>"` is the mandatory GitHub secret which guarantees that
-   hooks are coming from your GitHub repository and not a malfeasant.
+
+  * `match "<regex>" { <match-options> }` where `regex` is a regular expression
+    in [Rust regex format](https://docs.rs/regex/1.3.3/regex/index.html) that
```
vext01: We are linking to a very specific version of the regex crate docs here. I wonder if we should link to somepleace that will move with the times?

ltratt: Good point. Fixed in https://github.com/softdevteam/snare/pull/13/commits/18c83e10a049c2f1c5cd76fb0dfe7032d3ca22a4.

src/config.rs
```diff
@@ -230,10 +308,8 @@ fn search_snare_conf() -> Option<PathBuf> {
...
 }
 
-pub enum GenericOption<StorageT> {
-    Email(Lexeme<StorageT>),
-    MaxJobs(Lexeme<StorageT>),
-    Port(Lexeme<StorageT>),
-    ReposDir(Lexeme<StorageT>),
-    Secret(Lexeme<StorageT>),
+/// The configuration for a given repository.
+pub struct RConfig<'a> {
```
vext01: How about `RepoConfig` for this type name?

ltratt: That's what I started off with, but the variable names all ended up being `rconf` so I shortened it. I suppose I could lengthen it again.

src/config.rs
```diff
@@ -162,19 +122,137 @@ impl Config {
...
+pub struct Match {
+    /// The regular expression to match against full owner/repo names.
+    re: Regex,
+    /// An optional email address to send errors to.
+    email: Option<String>,
+    /// The GitHub secret used to validate requests.
+    secret: Option<SecStr>,
+}
+
+impl Default for Match {
```
vext01: Clever! ;)

src/config.rs
```diff
@@ -230,10 +308,8 @@ fn search_snare_conf() -> Option<PathBuf> {
...
 
-pub enum GenericOption<StorageT> {
-    Email(Lexeme<StorageT>),
-    MaxJobs(Lexeme<StorageT>),
-    Port(Lexeme<StorageT>),
-    ReposDir(Lexeme<StorageT>),
-    Secret(Lexeme<StorageT>),
+/// The configuration for a given repository.
+pub struct RConfig<'a> {
+    pub email: Option<&'a str>,
```
vext01: Who owns the data behind these references? The parser/lexer?

src/config.rs
ltratt: Fixed in https://github.com/softdevteam/snare/pull/13/commits/32c4d2884fa532b6f5f0cca59d2eb21dfd10add8.

src/config.rs
ltratt: The config.

src/config.rs
vext01: Thanks. That's much clearer.

I'd have underscored the non-camel-case variable names, but won't insist.

ltratt:
Squashed.

PR #15, opened 2020-01-30
Reload config after sighup

ltratt:
This PR enables snare to reload its config on SIGHUP. This is a bit more involved than it first seems because snare is multi-threaded. This PR thus goes through several stages (putting `Config` behind a `Mutex` and so on), before actually handling SIGHUP. Note that one case is handled somewhat, but not perfectly: reducing the value of `maxjobs`. I could do more here, but I think this is a niche case, and it's hard to test: the basic approach in this PR seems decent enough to me for the time being.

### DIFF
--- a/src/jobrunner.rs
+++ b/src/jobrunner.rs
@@ -27,6 +27,9 @@ const WAIT_TIMEOUT: i32 = 1;
 
 struct JobRunner {
     snare: Arc<Snare>,
+    /// The maximum number of jobs we will run at any one point. Note that this may not necessarily
+    /// be the same value as snare.config.maxjobs.
+    maxjobs: usize,
     /// The running jobs, with `0..num_running` entries.
     running: Vec<Option<Job>>,
     /// How many `Some` entries are there in `self.running`?
@@ -40,15 +43,16 @@ struct JobRunner {
 
 impl JobRunner {
     fn new(snare: Arc<Snare>) -> Result<Self, Box<dyn Error>> {
-        assert!(snare.config.maxjobs < std::usize::MAX);
-        let mut running = Vec::with_capacity(snare.config.maxjobs);
-        running.resize_with(snare.config.maxjobs, || None);
-        let mut pollfds = Vec::with_capacity(snare.config.maxjobs * 2 + 1);
-        pollfds.resize_with(snare.config.maxjobs * 2 + 1, || {
-            PollFd::new(-1, PollFlags::empty())
-        });
+        // If the unwrap() on the lock fails, the other thread has paniced.
+        let maxjobs = snare.config.lock().unwrap().maxjobs;
+        assert!(maxjobs < std::usize::MAX);
+        let mut running = Vec::with_capacity(maxjobs);
+        running.resize_with(maxjobs, || None);
+        let mut pollfds = Vec::with_capacity(maxjobs * 2 + 1);
+        pollfds.resize_with(maxjobs * 2 + 1, || PollFd::new(-1, PollFlags::empty()));
         Ok(JobRunner {
             snare,
+            maxjobs,
             running,
             num_running: 0,
             pollfds,
@@ -71,9 +75,7 @@ impl JobRunner {
             // able to run for temporary reasons, then wait a short amount of time and try again.
             // Notice that the second clause is a bit subtle: if there are jobs on the queue, but
             // we're all running the maximum number of jobs, then there's no point in waking up.
-            let timeout = if num_waiting > 0
-                || (check_queue && self.num_running < self.snare.config.maxjobs)
-            {
+            let timeout = if num_waiting > 0 || (check_queue && self.num_running < self.maxjobs) {
                 WAIT_TIMEOUT * 1000
             } else {
                 -1
@@ -83,7 +85,7 @@ impl JobRunner {
             // See if any of our active jobs have events. Knowing when a pipe is actually closed is
             // surprisingly hard. https://www.greenend.org.uk/rjk/tech/poll.html has an interesting
             // suggestion which we adapt slightly here.
-            for i in 0..self.snare.config.maxjobs {
+            for i in 0..self.maxjobs {
                 // stderr
                 if let Some(flags) = self.pollfds[i * 2].revents() {
                     if flags.contains(PollFlags::POLLIN) {
@@ -175,7 +177,7 @@ impl JobRunner {
             }
 
             // Has the HTTP server told us that it's put more jobs into the queue?
-            match self.pollfds[self.snare.config.maxjobs * 2].revents() {
+            match self.pollfds[self.maxjobs * 2].revents() {
                 Some(flags) if flags == PollFlags::POLLIN => {
                     check_queue = true;
                     // It's fine for us to drain the event pipe completely: we'll process all the
@@ -194,7 +196,7 @@ impl JobRunner {
             // it fully, or because the HTTP server has told us that it's put more jobs there.
             // However, it's only worth us checking the queue (which requires a lock) if there's
             // space for us to run further jobs.
-            if check_queue && self.num_running < self.snare.config.maxjobs {
+            if check_queue && self.num_running < self.maxjobs {
                 check_queue = !self.try_pop_queue();
             }
         }
@@ -211,7 +213,7 @@ impl JobRunner {
             let pjob = self.snare.queue.lock().unwrap().pop();
             match pjob {
                 Some(qj) => {
-                    if self.num_running == self.snare.config.maxjobs {
+                    if self.num_running == self.maxjobs {
                         self.snare.queue.lock().unwrap().push_front(qj);
                         return false;
                     }
@@ -339,8 +341,7 @@ impl JobRunner {
             self.pollfds[i * 2] = PollFd::new(stderr_fd, PollFlags::POLLIN);
             self.pollfds[i * 2 + 1] = PollFd::new(stdout_fd, PollFlags::POLLIN);
         }
-        self.pollfds[self.snare.config.maxjobs * 2] =
-            PollFd::new(self.snare.event_read_fd, PollFlags::POLLIN);
+        self.pollfds[self.maxjobs * 2] = PollFd::new(self.snare.event_read_fd, PollFlags::POLLIN);
     }
 
     /// If the user has specified an email address, send the contents of
--- a/src/jobrunner.rs
+++ b/src/jobrunner.rs
@@ -82,10 +82,7 @@ impl JobRunner {
             };
             poll(&mut self.pollfds, timeout).ok();
 
-            // The SIGHUP listener writes to the event FD if SIGHUP has been received, so it's
-            // possible that the reason we've been woken up is that the SIGHUP event needs to be
-            // processed.
-            self.snare.check_for_hup();
+            self.check_for_sighup();
 
             // See if any of our active jobs have events. Knowing when a pipe is actually closed is
             // surprisingly hard. https://www.greenend.org.uk/rjk/tech/poll.html has an interesting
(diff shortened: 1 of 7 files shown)


vext01:
> putting Config behind a Mutex and so on

Is the config mutable after parsing then? That would surprise me.

ltratt:
> Is the config mutable after parsing then? That would surprise me.

When you send SIGHUP, the entire config file is reparsed and a new `Config` produced. So the `Config` is not mutable as such, but it can be replaced by a new `Config`.

vext01:
> When you send SIGHUP, the entire config file is reparsed and a new Config produced. So the Config is not mutable as such, but it can be replaced by a new Config.

In the past, I've just had the program re-`exec(3)` itself, then you don't have to worry about any of this. I'm not sure if that would work for us here?

ltratt:
> In the past, I've just had the program re-exec(3) itself

That would not be good here since we'd destroy the queue of jobs!

vext01:
> That would not be good here since we'd destroy the queue of jobs!

Well, you'd have to wait for them to finish. I thought you were doing that already, but I guess from your response that you allow them to continue during the reload?

ltratt:
Yes, the reason this PR is quite so fiddly is that it has no effect on ongoing jobs, but we try still try to enact all the reasonable changes as soon as possible.

vext01:
I see. I hadn't appreciated that!

Code review coming soon.

ltratt:
Simplifying a bit: we don't change the config of any running job. So if a job was run when you said `email="a@b"` and you SIGHUP it so `email="c@d"` then the existing job will still send to `a@b` but *all* new jobs will send to `c@d`. As this suggests, each job has its own "unique" config in a sense (`RepoConfig`).

The major exception is https://github.com/softdevteam/snare/pull/15/commits/82af641c98f2a2b62fa9e153514694eb7371e16e: it's really fiddly -- and in the general case impossible -- to deal well with SIGHUP reducing the number of `maxjobs` if there are already running jobs. So I implemented something simple, given that this is a fairly niche case https://github.com/softdevteam/snare/pull/15/commits/82af641c98f2a2b62fa9e153514694eb7371e16e#diff-b319aab93ab499624a467ced0e18a2a8R364.

vext01:
> So if a job was run when you said email="a@b" and you SIGHUP it so email="c@d" then the existing job will still send to a@b

I think that's OK. As I user I'd kind of expect that.

> to deal well with SIGHUP reducing the number of maxjobs if there are already running jobs. 

Hrm, yes. That's annoying.

I can think of two "possible improvements":

 - Wait until there are fewer-or-equal jobs than the new maxjobs, then create a new smaller array from the used slot of the old array. I think this is what your comment suggests with "compaction".

 - Keep a count of how many jobs there are, but the backing array may be larger. May waste memory and annoying to keep in sync.

I think the solution you have now is pretty good tbh.

src/jobrunner.rs
```diff
@@ -40,15 +43,16 @@ struct JobRunner {
...
 impl JobRunner {
     fn new(snare: Arc<Snare>) -> Result<Self, Box<dyn Error>> {
-        assert!(snare.config.maxjobs < std::usize::MAX);
-        let mut running = Vec::with_capacity(snare.config.maxjobs);
-        running.resize_with(snare.config.maxjobs, || None);
-        let mut pollfds = Vec::with_capacity(snare.config.maxjobs * 2 + 1);
-        pollfds.resize_with(snare.config.maxjobs * 2 + 1, || {
-            PollFd::new(-1, PollFlags::empty())
-        });
+        // If the unwrap() on the lock fails, the other thread has paniced.
```
vext01: Is it necessary to repeat this comment on every mutex unlock? Mutex poisoning is well-known among Rust programmers and the documentation explains it well.

src/jobrunner.rs
```diff
@@ -40,15 +43,16 @@ struct JobRunner {
...
+        let maxjobs = snare.config.lock().unwrap().maxjobs;
+        assert!(maxjobs < std::usize::MAX);
```
vext01: I'm not sure this is a meaningful assertion. Since both arguments are `usize` it's equivalent to:
```
assert!(maxjobs != std::usize::MAX);
```

Is that being used as a boundary condition elsewhere perhaps?

src/jobrunner.rs
ltratt: I tend to agree. I think in a future PR I will put this on the attribute in the struct.

src/jobrunner.rs
```diff
@@ -40,15 +43,16 @@ struct JobRunner {
...
+        let mut running = Vec::with_capacity(maxjobs);
+        running.resize_with(maxjobs, || None);
+        let mut pollfds = Vec::with_capacity(maxjobs * 2 + 1);
```
vext01: Perhaps the above assertion is to cater for the `+ 1` here?

src/main.rs
```diff
@@ -37,10 +41,34 @@ use queue::Queue;
...
     event_write_fd: RawFd,
+    /// Has a SIGHUP event occurred? If so, the jobrunner will process it, and set this to false in
+    /// case future SIGHUP events are detected.
+    sighup_occurred: Arc<AtomicBool>,
+}
+
+impl Snare {
+    /// Check to see if we've received a SIGHUP since the last check. If so, we will reload the
+    /// config file. **Note that because snare has multiple threads, the config file can change at
+    /// any arbitrary point, not just after calling this function.**
```
vext01: I don't understand the "not just after calling this function" part of the comment.

When else can the config change?

src/main.rs
```diff
@@ -37,10 +41,34 @@ use queue::Queue;
...
+    fn check_for_hup(&self) {
+        if self.sighup_occurred.load(Ordering::Release) {
```
vext01: Is the ordering supposed to be acquire here?

src/jobrunner.rs
ltratt: Actually, this should be `assert!(maxjobs < ((std::usize::MAX - 1) / 2)`! Long story, but it's basically about the amount of pollfds we create. In practise, of course, we're probably not going to have enough RAM for this to ever be an issue.

Fixed in https://github.com/softdevteam/snare/pull/15/commits/5ba7c259a8e0c501cdced7f99020ca1efe9bc2c1.

src/jobrunner.rs
ltratt: Correct.

src/main.rs
ltratt: There are two threads in snare, so the config can change in another thread even if it's not called `check_for_hup`.

src/main.rs
```diff
@@ -119,11 +147,27 @@ pub async fn main() {
     };
 
     let (event_read_fd, event_write_fd) = pipe2(OFlag::O_NONBLOCK).unwrap();
+    let sighup_occurred = Arc::new(AtomicBool::new(false));
+    {
+        let sighup_occurred = Arc::clone(&sighup_occurred);
+        if let Err(e) = unsafe {
```
vext01: If you stored the closure in a variable, I think you might be able to limit the scope of `unsafe` some more?

src/main.rs
ltratt: I think `Relaxed` is fine here as we don't need any other read/writes to have occurred before/after this.

src/main.rs
```diff
@@ -63,18 +63,23 @@ impl Snare {
...
     /// any arbitrary point, not just after calling this function.**
     fn check_for_hup(&self) {
-        if self.sighup_occurred.load(Ordering::Release) {
-            let config = Config::from_path(&self.conf_path);
-            *self.config.lock().unwrap() = config;
-            self.sighup_occurred.store(false, Ordering::Release);
+        if self.sighup_occurred.load(Ordering::Relaxed) {
+            match Config::from_path(&self.conf_path) {
+                Ok(config) => *self.config.lock().unwrap() = config,
+                Err(msg) => eprintln!("{}", msg),
```
vext01: So the error is printed on stderr, which may be in the background and may go un-noticed.

Not sure how you can fix that though.

One potential idea would be to have a `snare --reload` which communicates with the existing snare instance and prints any errors to its stderr, (not the daemon's). However, this would probably need more complex IPC.

src/main.rs
```diff
@@ -63,18 +63,23 @@ impl Snare {
...
 
 /// Exit with a fatal error.
 fn fatal(msg: &str) -> ! {
-    debug_assert!(msg.ends_with('.'));
-    eprintln!("{}.", msg);
+    if msg.ends_with('.') {
+        eprintln!("{}", msg);
+    } else {
+        eprintln!("{}.", msg);
+    }
```
vext01: I wonder if this is worth it :)

There's a any number of silly things the caller might do.

```
msg("error..");
msg("error,");
...
```

src/main.rs
ltratt: A later PR will send this to syslog.

src/jobrunner.rs
vext01: ok

src/jobrunner.rs
vext01: OK, that looks more valid at least.

I missed why the `/ 2`, but if the story is really long, I'll trust you ;)

src/main.rs
vext01: To be clear, the config may change when **neither** thread received a HUP?

src/main.rs
vext01: But we have `Release`.

Perhaps both this and the ordering in a few lines time should both be relaxed?

(If the config weren't in a mutex, you'd certainly want acquire/release, otherwise another thread may see the config mid-move)

src/jobrunner.rs
ltratt: Each job has stderr/stdout pipes which is the `*` (or `/)` 2.

src/main.rs
vext01: sounds good!

src/main.rs
ltratt: At the moment, the signal comes in, a global bool is set, and then `jobrunner` checks that bool and reloads the config if necessary. There are only two threads, so one of them has to handle it (and the config stuff is way too much for it to be signal safe, so the signal handler can't deal with it directly).

src/main.rs
ltratt: Yeah, I'm not sure either :/

src/jobrunner.rs
vext01: ok

src/main.rs
vext01: OK understood. But the comment makes it sound like some other part of the program (not this function) is mutating the config. As I understand you mean to say instead that this function could be run in another thread.

src/main.rs
ltratt: You're totally right: they should both be `Relaxed` and one of the later commits (https://github.com/softdevteam/snare/pull/15/commits/ff45ea1e03b8c4d86a38cd6c04e38e691c88c736#diff-639fbc4ef05b315af92b4d836c31b023R66) fixes that. I don't know why I ever thougt `Release` was the correct ordering!

src/main.rs
ltratt: I'm not sure I understand?

vext01: So if you did something like:
```
if let Err(e) = {
    let f = move || {
        // All functions called in this function must be signal safe. See signal(3).
        sighup_occurred.store(true, Ordering::Relaxed);
        unsafe { nix::unistd::write(event_write_fd, &[0]).ok() };
    };
    unsafe { signal_hook::register(signal_hook::SIGHUP, f) };
}
```

Then fewer lines can be in `unsafe`? I may be wrong.

src/main.rs
ltratt: Is https://github.com/softdevteam/snare/pull/15/commits/3f03f446c341e2b9b852f4f2998203f4ba7022d6 better?

src/main.rs
ltratt: TBH, I'm fairly happy having the closure in `unsafe` because it's a signal handler and being clear that "here be dragons" is not a bad idea.

src/main.rs
vext01: Much clearer, thanks.

ltratt:
I think that's everything?

vext01:
LGTM. Please squash.

ltratt:
Squashed.

ltratt:
@vext01 Any idea why this failed? buildbot seems to have succeeded but bors failed?

vext01:
Yeah, that doesn't look like our issue.

Let's try again.

bors r+

vext01:
once more for luck

bors r+

PR #16, opened 2020-01-31
Add a timeout facility to jobs.

ltratt:
If `maxjobs` number of jobs get stuck (e.g. waiting to download something), snare will no longer be able to make progress. A simple timeout is the only way I can see to ensure that progress does occur (even if the progress is slow).


src/jobrunner.rs
```diff
@@ -140,10 +162,26 @@ impl JobRunner {
...
             num_waiting = 0;
+            next_finish_by = None;
             for i in 0..self.running.len() {
+                if let Some(Job {
+                    finish_by,
+                    ref child,
+                    ..
+                }) = self.running[i]
+                {
+                    if finish_by <= Instant::now() {
```
vext01: So does this rely on the event loop waking up?

Would the code still work if we had maxjobs jobs which all hang?

src/jobrunner.rs
```diff
@@ -304,7 +342,19 @@ impl JobRunner {
                             return Err(None);
                         }
 
+                        // This unwrap() is, in theory, unsafe because we could exceed the timeout
+                        // duration. However, a quick back-of-the-envelope calculation suggests
+                        // that, assuming `Instant` is a `u64`, this could only happen with an
+                        // uptime of over 500,000,000 years. This seems adequately long that I'm
+                        // happy to take the risk on the unwrap().
```
vext01: This comment amuses me :)

vext01:
There are some changes in here which are not to do with timeouts, e.g. `very` -> `sec`.

I won't insist, but they should have been in separate commits.

src/jobrunner.rs
ltratt: This makes `poll` wake up in time to check the next timeout.

ltratt:
> There are some changes in here which are not to do with timeouts, e.g. very -> sec.

I think you're being overly pedantic here. The reason that changed is because the example below it would not have made as much sense otherwise.

src/jobrunner.rs
vext01: jolly good :)

vext01:
Fair enough.

So ready to merge?

ltratt:
Yep, but if bors doesn't work, we can try waiting a day or so, as I don't have anything else after this at the moment.

vext01:
I'm manually merging this because bors is still playing up.

PR #17, opened 2020-01-31
Queue kinds

ltratt:
This PR adds queue kinds: `queue = <evict|parallel|sequential>`. Previously `parallel` was all we supported, which had some comic outcomes: for example, for yksom, you could get race conditions trying to force push the `gh-pages` branch. `sequential` is now the default. Hopefully the README explains why these exist and why they might be useful.

This is the last major feature I think we need before a 0.1 release (though there are a few minor bits that probably need doing).

### DIFF
--- a/src/jobrunner.rs
+++ b/src/jobrunner.rs
@@ -245,22 +245,21 @@ impl JobRunner {
     /// Try to pop all jobs on the queue: returns `true` if it was able to do so successfully or
     /// `false` otherwise.
     fn try_pop_queue(&mut self) -> bool {
+        let snare = Arc::clone(&self.snare);
+        let mut queue = snare.queue.lock().unwrap();
         loop {
-            let pjob = {
-                let mut queue = self.snare.queue.lock().unwrap();
-                if self.num_running == self.maxjobs && !queue.is_empty() {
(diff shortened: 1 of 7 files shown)


src/jobrunner.rs
```diff
@@ -235,7 +235,7 @@ impl JobRunner {
             // Should we check the queue? This could be because we were previously unable to empty
             // it fully, or because the HTTP server has told us that it's put more jobs there.
             // However, it's only worth us checking the queue (which requires a lock) if there's
-            // space for us to run further jobs.
+            // space for us to run further jobs. Note that this check *must* come after
```
vext01: Is that a partial comment, or does it just need a full-stop?

ltratt: Oops, I got half way through writing this, and realised I could solve the problem another way. So this is now misleading. Removed in https://github.com/softdevteam/snare/pull/17/commits/caadd495cdf6ad95ca0a07bd3a844da0122d7f11.

vext01:
This LGTM.

Please squash and we can test if bors is feeling better today.

ltratt:
Squashed.

PR #18, opened 2020-02-01
Top level options

ltratt:
This PR adds two top-level options: `user` and `listen` (the latter replacing `port`). The two commits hopefully make fairly clear what these do and why (including the current limitations of `listen`).


src/main.rs
```diff
@@ -110,6 +113,46 @@ fn search_snare_conf() -> Option<PathBuf> {
     None
 }
 
+/// If the config specified a 'user' then switch to that and update $HOME and $USER appropriately.
+fn change_user(conf: &Config) {
+    match conf.user {
+        Some(ref user) => match get_user_by_name(&user) {
```
vext01: It hadn't occurred to me before, but looking up users offers racy scenarios.

For example here, you look up a user by name, getting back a `User` struct containing numeric IDs etc, but that struct could go stale at any moment.

I don't there's a way to fix this though(?)

ltratt: It's an inevitable issue -- there is no global lock on the OS and user! [Imagine if there was -- your computer would have to ban you from working at some points.]

vext01: Yes. I also concluded that you'd need some kernel change, but a lock would be madness. Perhaps some kind of compare and swap scenario, but now we are off on a tangent.

ltratt: It's absolutely impossible -- you'd have to stop the user doing anything! It would be the lock to end all locks ;)

vext01:
There's a typo in the first line of the second commit message.

And:

> Unfortunately I'm not sure that hyper can listen on more than one address at a
time

I encountered the same problem when I was messing around with tokio last year. The good news is needing multiple bind addresses is fairly rare and you can use a firewall forward rule to work around it.

So aside from the typo. This LGTM.

ltratt:
Do you want me to force push an update?

vext01:
Yes please.

ltratt:
Updated.

PR #19, opened 2020-02-01
Change the order that we search for snare.conf.

ltratt:
It makes more sense to search for the per-user .snare.conf first before falling back to trying the global /etc/snare.conf.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,8 +17,8 @@ where:
 
  * `<config-path>` is a path to a `snare.conf` configuration file. If not
    specified explicitly, the following locations will be searched, in order:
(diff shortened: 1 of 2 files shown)


vext01:
> It makes more sense to search for the per-user .snare.conf first before falling back to trying the global /etc/snare.conf.

I agree, otherwise unprivileged users may not be able to run their own snare instances.

By the way, did you think about using XDG-style paths (`~/.config/snare`)? Personally I don't really like it, but some people do.

ltratt:
snare is a proper old-school-beardy Unix daemon. None of that `~/.config` modern stuff.

PR #20, opened 2020-02-03
Daemonise by default (unless -d is passed on the command line).

ltratt:
This requires a bit of reshuffling of the main method, as something in the `[tokio::main]` setup doesn't work with `daemon` (perhaps it creates threads, as `fork` only carries the main thread over to the child process).

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -24,6 +24,8 @@ lrlex_mod!("config.l");
 lrpar_mod!("config.y");
 
 pub struct Config {
+    /// If true, then daemonise (i.e. disassociate from the terminal and log errors via syslog).
+    pub daemonise: bool,
     /// The IP address/port on which to listen.
     pub listen: SocketAddr,
     /// The maximum number of parallel jobs to run.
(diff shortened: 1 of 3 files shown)


vext01:
So you didn't have to do anything special to get output to go to syslog?

ltratt:
This PR has nothing to say about syslog, but syslog will be the topic of the next PR after this...

src/config.rs
```diff
@@ -24,6 +24,8 @@ lrlex_mod!("config.l");
 lrpar_mod!("config.y");
 
 pub struct Config {
+    /// If true, then daemonise (i.e. disassociate from the terminal and log errors via syslog).
```
vext01: Because it says here that errors will go to syslog. Perhaps drop that part until the support is implemented?

ltratt: The support is implemented, but I plonked it in a separate PR. I'm inclined to leave this comment as-is -- it's a minor, and very temporary, wart.

vext01: OK

PR #21, opened 2020-02-03
Log via syslog after daemonisation

ltratt:


src/main.rs
```diff
@@ -197,6 +250,22 @@ pub fn main() {
...
     }
+    let progname = match current_exe() {
+        Ok(p) => p
+            .file_name()
+            .map(|x| x.to_str().unwrap_or("snare"))
+            .unwrap_or("snare")
+            .to_owned(),
+        Err(_) => "snare".to_owned(),
+    };
+    // openlog's first argument `ident` is incompletely specified, but in practise we have to assume that
```
vext01: I don't think it's necessary to leak memory. You just need to ensure that the `CString` lives long enough to see the last call to `syslog()`, right?

Sadly `CString::new` isn't const, otherwise you could use a global constant, but you could use a `lazy_static!`?

How many times do you leak? If it's at most one, then what you have is probably OK.

vext01:
Just one query.

src/main.rs
ltratt: We only leak once. We don't ever call `closelog` because, in theory, `snare` should never stop; so it's arguably not worth being terribly clever about this.

vext01: Fair enough. I can live with it.

PR #24, opened 2020-02-03
Respond to GitHub pings.

ltratt:
When a user installs a webhook, GitHub pings the webhook server. This simple commit responds with "success" if, and only if, authentication succeeds *and* there is a script for the repository in question.

@vext01 There is an interesting question here. Should we return success after authentication, even if there's not a script to run? To my mind, that equates to "things aren't fully set up yet, so best to make it clear", but I could be convinced otherwise.

### DIFF
--- a/src/httpserver.rs
+++ b/src/httpserver.rs
@@ -66,6 +66,11 @@ async fn handle(req: Request<Body>, snare: Arc<Snare>) -> Result<Response<Body>,
         return Ok(res);
     }
 
+    if event_type == "ping" {
+        *res.status_mut() = StatusCode::OK;
+        return Ok(res);
+    }
+
     // We now check that there is a per-repo program for this repository and that we haven't been
     // tricked into searching for a file outside of the repos dir.
     let mut p = PathBuf::new();


vext01:
My gut feeling is that we should return success if there are no scripts yet. Then the user can set up in stages and can test network reachability aside from the scripts themselves.

(You may even warn on syslog if there are no scripts for a repo?)

But I don't feel all that strongly, if you'd rather do something else.

ltratt:
> My gut feeling is that we should return success if there are no scripts yet. Then the user can set up in stages and can test network reachability aside from the scripts themselves.

OK, I can buy this.

> You may even warn on syslog if there are no scripts for a repo?

Yes, I was thinking of doing that anyway.

ltratt:
https://github.com/softdevteam/snare/pull/24/commits/8c31e2239a667d49d68c8547bb0cfdd1a7598a7d responds to any ping once authentication is successful. [The syslog thing I think should be a separate PR.] Work for you?

vext01:
LGTM, please squash.

ltratt:
Squashed.

PR #26, opened 2020-02-03
Fail authentication if a secret is present.

ltratt:
Previously we authenticated successfully if a signature was present, but if the GitHub side shared a secret, and our side didn't, we let it through anyway. In one sense this is fine, but it means that one can easily forget to specify a secret (and when I say "one", I mean "I have already made this mistake").

This commit thus returns `UNAUTHORIZED` if: the request is signed and we don't have a secret; we have a secret and the request isn't signed; or our secret doesn't match the signed request.

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,7 +85,10 @@ A `match` block supports the following options:
        jobs will stay on the queue and be executed in FIFO order.
  * `secret = "<secret>";` is an optional GitHub secret which guarantees that
    hooks are coming from your GitHub repository and not a malfeasant. Although
-   this is optional, we *highly* recommend setting it in all cases.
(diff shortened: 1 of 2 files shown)


vext01:
LGTM, but can we add a note in the user documentation somewhere?

ltratt:
> LGTM, but can we add a note in the user documentation somewhere?

Good idea. Added in https://github.com/softdevteam/snare/pull/26/commits/1dba0e389af2bfbc49e43da1a1c17b7b87697e51

vext01:
Please squash.

ltratt:
Squashed.

PR #28, opened 2020-02-05
Audit all the unwrap() calls.

ltratt:
In each case do one of the following:

  * Ensure they are documented (and document if not) to explain why they're     safe.
  * If they're not safe, make them safe (e.g. add a fallback, log an error).

The aim here is that snare should not fail at runtime due to `unwrap()` calls.

### DIFF
--- a/src/jobrunner.rs
+++ b/src/jobrunner.rs
@@ -48,7 +48,6 @@ struct JobRunner {
 
 impl JobRunner {
     fn new(snare: Arc<Snare>) -> Result<Self, Box<dyn Error>> {
-        // If the unwrap() on the lock fails, the other thread has paniced.
         let maxjobs = snare.conf.lock().unwrap().maxjobs;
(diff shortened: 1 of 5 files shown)


src/jobrunner.rs
```diff
@@ -109,6 +108,12 @@ impl JobRunner {
             // See if any of our active jobs have events. Knowing when a pipe is actually closed is
             // surprisingly hard. https://www.greenend.org.uk/rjk/tech/poll.html has an interesting
             // suggestion which we adapt slightly here.
+            //
+            // This `for` loop has various unwrap() calsl. If `flags[i * 2]` or `flags[i * 2 + 1]`
```
vext01: `calsl`

ltratt: Fixed in https://github.com/softdevteam/snare/pull/28/commits/ce102a545852a819d497cc57063a68db3548e347.

vext01:
Please squash

ltratt:
Squashed.

PR #30, opened 2020-02-05
Support for cargo-deny

ltratt:
This PR adds support for [cargo-deny](https://embarkstudios.github.io/cargo-deny/index.html), which makes it harder for us (or a dependency) to accidentally undermine the license we have chosen.

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -1,0 +1,16 @@
+[licenses]
+unlicensed = "deny"
+confidence-threshold = 1.0
+allow = [
+    "Apache-2.0",
+    "BSD-3-Clause",
+    "MIT",
+    "Unlicense"
+]
+
+# We'd prefer not to use the MPL if possible, so explicitly whitelist it each
+# time we have to use it.
+exceptions = [
+    { name = "ascii_utils", allow = ["MPL-2.0"] },
+    { name = "fast_chemail", allow = ["MPL-2.0"] }
+]
--- a/.buildbot.sh
+++ b/.buildbot.sh
@@ -12,3 +12,6 @@ export PATH=`pwd`/.cargo/bin/:$PATH
 
 cargo fmt --all -- --check
 cargo build
+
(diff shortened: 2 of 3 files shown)


ltratt:
bors try

ltratt:
Can we sensible install `cargo-deny` somewhere so that we don't waste energy building it anew on each PR?

deny.toml
```diff
@@ -0,0 +1,14 @@
...
+unlicensed = "deny"
+confidence-threshold = 1.0
+allow = [
+    "Apache-2.0",
+    "BSD-3-Clause",
+    "MIT",
+    "Unlicense"
+]
+
+exceptions = [
```
vext01: How comes these exceptions?

vext01:
> Can we sensibly install cargo-deny somewhere so that we don't waste energy building it anew on each PR?

We can, but for such a short build it's probably better to build the latest version every time isn't it?

deny.toml
ltratt: The exact status of the MPL is a bit fishy (in the sense that it's less clear than other licenses about its virality), and we really don't want to allow any more such code in unless we absolutely have to. Doing things this way means that we have to explicitly allow the MPL in each time.

ltratt:
> We can, but for such a short build it's probably better to build the latest version every time isn't it?

For this PR I guess we can leave things as-is, but I am keen that in the future we don't consume energy unnecessarily: compiling has consequences!

deny.toml
vext01: OK. So do we need to add a comment to that effect?

ltratt: Point taken. Fixed in https://github.com/softdevteam/snare/pull/30/commits/92489ab5b56daf2cd1efde0c5c437003332b745c.

vext01:
Please squash.

ltratt:
Squashed.

PR #35, opened 2020-02-12
Add install instructions and a basic Makefile.

ltratt:
In one sense the Makefile is unnecessary, but it means that normal Unix people who aren't familiar with Rust simply need to install Rust and then type "make install" in order for something sensible to happen. That feels like a reasonable compromise between old and new to me.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,17 @@
+PREFIX ?= /usr/local
+MAN_PREFIX ?= ${PREFIX}/man
+
+all:
+	cargo build --release
+
+install:
+	cargo build --release
+	install -d ${PREFIX}/bin
+	install -c -m 555 target/release/snare ${PREFIX}/bin/snare
+	install -d ${MAN_PREFIX}/man/man1
+	install -d ${MAN_PREFIX}/man/man5
+	install -c -m 444 snare.1 ${MAN_PREFIX}/man/man1/snare.1
+	install -c -m 444 snare.conf.5 ${MAN_PREFIX}/man/man5/snare.conf.5
+	install -d ${PREFIX}/share/examples/snare
+	install -c -m 444 snare.conf.example ${PREFIX}/share/examples/snare
+
--- a/README.md
+++ b/README.md
@@ -5,6 +5,15 @@ arbitrary program for that repository informing it of the event -- that program
 can then perform whatever actions it wants.
 
 
+## Install
+
+You will need rustc-1.40.0 or greater installed. You may use `cargo` to build
+snare locally or you may use the `Makefile` to build and install `snare` in
+traditional Unix fashion. `make install` defaults to installing in
+`/usr/local`: you can override this by setting the `PREFIX` variable to another
+path (e.g. `PREFIX=/opt/local make`).
+
+
 ## Quick setup
 
 `snare` has the following command-line format:


vext01:
Since cargo [cannot yet install auxiliary files](https://github.com/rust-lang/cargo/issues/2729), I think this is the way to go.

ltratt:
OK, does that mean I need to change anything or do you want to bors it or ...?

Makefile
```diff
@@ -0,0 +1,15 @@
+PREFIX ?= /usr/local
+
+all:
+	cargo build --release
+
+install:
+	cargo build --release
+	install -d ${PREFIX}/bin
+	install -c -m 555 target/release/snare ${PREFIX}/bin/snare
+	install -d ${PREFIX}/man/man1
```
vext01: These manual paths are only correct for OpenBSD. In Linux you'd install into `${PREFIX}/share/man`.

Is it worth fixing? Maybe not.

vext01:
Did you try with gmake too?

ltratt:
> Did you try with gmake too?

Yep -- it works fine.

vext01:
So just that one comment about paths.

Makefile
ltratt: I don't see a sensible way of making this fully generic without going down the autoconf path. To be honest, anyone who notices that will probably be capable of editing the `Makefile` paths.

vext01: I good and easy compromise is to add a `MAN_PREFIX` variable like you did for `PREFIX`.

vext01: Or you can check the OS in the makefile without autoconf. Admittedly, that's nicer/easier for gmake.

ltratt: OK, I think https://github.com/softdevteam/snare/pull/35/commits/92592c5604598f43aa3e8c13a5348fe21cfb3241 might be the trick?

Makefile
```diff
@@ -0,0 +1,17 @@
...
+	cargo build --release
+	install -d ${PREFIX}/bin
+	install -c -m 555 target/release/snare ${PREFIX}/bin/snare
+	install -d ${MAN_PREFIX}/man/man1
+	install -d ${MAN_PREFIX}/man/man5
+	install -c -m 444 snare.1 ${MAN_PREFIX}/man/man1/snare.1
+	install -c -m 444 snare.conf.5 ${MAN_PREFIX}/man/man5/snare.conf.5
+	install -d ${PREFIX}/share/examples/snare
+	install -c -m 444 snare.conf.example ${PREFIX}/share/examples/snare
+
```
vext01: Was this trailing whitespace intentional?

Makefile
vext01: That'll do the trick!

Makefile
ltratt: Yes, I should have put it in there at first. Old-fashioned Unix files are supposed to have a blank last line. I normally ignore this, but since this file is for old-fashioned Unix purposes, it seemed appropriate.

vext01: OK, I didn't realise github rendered them. Or at least, usually it puts a funny symbol to show the absence of a trailing newline.

Anyway...

vext01:
Please squash.

ltratt:
Squashed.

PR #39, opened 2020-02-19
Validate input

ltratt:
This PR validates the GitHub user and repository names we receive in a request to ensure a) that they're valid b) that when included in a pathname they can't accidentally cause us to reference a file we shouldn't have access to.

### DIFF
--- a/src/httpserver.rs
+++ b/src/httpserver.rs
@@ -60,6 +60,11 @@ async fn handle(req: Request<Body>, snare: Arc<Snare>) -> Result<Response<Body>,
         }
     };
 
+    if !valid_github_username(&owner) {
+        *res.status_mut() = StatusCode::BAD_REQUEST;
+        return Ok(res);
+    }
+
     let conf = snare.conf.lock().unwrap();
     let (rconf, secret) = conf.github.repoconfig(&owner, &repo);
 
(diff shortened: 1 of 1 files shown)


src/httpserver.rs
```diff
@@ -209,6 +209,23 @@ fn valid_github_username(n: &str) -> bool {
     n.chars().all(|c| c.is_ascii_alphanumeric() || c == '-')
 }
 
+/// Is `n` a valid GitHub repository name?
```
vext01: Did you get these rules by supplying an invalid reponame as you did for user names?

src/httpserver.rs
```diff
@@ -97,24 +97,23 @@ async fn handle(req: Request<Body>, snare: Arc<Snare>) -> Result<Response<Body>,
         return Ok(res);
     }
 
-    // We now check that there is a per-repo program for this repository and that we haven't been
-    // tricked into searching for a file outside of the repos dir.
+    // We now check that there is a per-repo program for this repository.
     let mut p = PathBuf::new();
     p.push(&conf.github.reposdir);
+    // The calls to github_valid_username and github_valid_reponame above guarantee that `owner`
+    // and `repo` are safe to put in pathnames.
```
vext01: Would it also be worthwhile to escape user and repo names beginning with `~` as this expands to user homes. What about `$` in case the user can expand shell variables.

A possibly valid counter-example might be: we never (knowingly) pass these characters to a shell(?)

src/httpserver.rs
ltratt: Yep, same process. I should probably add a comment in to that effect. Fixed in https://github.com/softdevteam/snare/pull/39/commits/3c177aa7433e8a0a6576ffa12ae25f4d70d0bce6.

src/httpserver.rs
ltratt: `~` and `$` aren't valid characters, so they'll be rejected. Fortunately (and probably not coincidentally), GitHub's rules on this are onerous enough that nothing nasty can leak through.

ltratt: [The tests prove that these are rejected BTW: notice the `for i in 0..255` bits.]

vext01: Ah ok, good!

vext01:
Please squash

ltratt:
Squashed.

PR #41, opened 2020-02-19
More input validation

ltratt:
This PR tidies up a minor mistake from #39 (https://github.com/softdevteam/snare/commit/13b0b4af3c8583bea3c9a80517c3f681e3170b98), then validates the GitHub 'event' type (https://github.com/softdevteam/snare/commit/a9e1005e8ce731696d0b367cb8f57572fdd3441c) before explicitly logging why validation has failed (https://github.com/softdevteam/snare/commit/abdbe8268a46350fda60b9394a4df168af91e223).

### DIFF
--- a/src/httpserver.rs
+++ b/src/httpserver.rs
@@ -60,7 +60,7 @@ async fn handle(req: Request<Body>, snare: Arc<Snare>) -> Result<Response<Body>,
         }
     };
 
-    if !valid_github_username(&owner) || !valid_github_reponame(&owner) {
+    if !valid_github_ownername(&owner) || !valid_github_reponame(&owner) {
         *res.status_mut() = StatusCode::BAD_REQUEST;
         return Ok(res);
     }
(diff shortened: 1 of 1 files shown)


src/httpserver.rs
```diff
@@ -179,6 +179,13 @@ async fn parse(req: Request<Body>) -> Result<(Bytes, String, String, String), ()
     }
 }
 
+/// Is `t` a valid GitHub event type? If this function returns `true` then it is guaranteed that `t`
+/// is safe to use in pathnames.
```
vext01: `pathnames` -> `filesystem paths`?

So the event name is used later to build a path?

vext01:
Just one small comment.

src/httpserver.rs
ltratt: Amusingly, I originally had "filesystem" in there, before realising that I was really just saying the same thing twice, and taking it out. I think it's pretty clear in this case that we could only possibly be talking about filesystems.

vext01: Well, for a while I was confused as to whether you meant a part of the URL, before about 90% convincing myself you meant on the filesystem.

So if not `filesystem` then at least `path names` (two words).

ltratt: I take your point. Fixed in https://github.com/softdevteam/snare/pull/41/commits/8ced8f0b30c3f4dcc5e51e00c6bb0613b89a1c8d.

vext01:
Please squash

ltratt:
On reflection, I just needed to edit a commit message rather than squash. Ready for merging!

PR #43, opened 2020-02-24
Replace GitHub-level 'reposdir' with match-level 'cmd'.

ltratt:
This commit removes the `github`-block level `reposdir` and replaces it with the more flexible `match`-block level `cmd` has been introduced. In essence:

```
  github {
    reposdir = "/path/to/prps";
    ...
  }
```

should be changed to:

```
  github {
    match ".*" {
      cmd = "/path/to/reposdir/%o/%r %e %j";
    }
  }
```

in order to obtain the previous behaviour. `snare` informs users whose config contains `repodir` how to update it, hopefully minimising porting from snare-0.1.0.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,38 @@
+# snare 0.2.0 (xxxx-xx-xx)
+
+## Breaking changes
+
+* The `github`-block level `reposdir` option has been removed. The more
+  flexible `match`-block level `cmd` has been introduced. In essence:
+
+    ```
+    github {
+      reposdir = "/path/to/prps";
+      ...
+    }
+    ```
+
+  should be changed to:
+
+    ```
+    github {
+      match ".*" {
+        cmd = "/path/to/reposdir/%o/%r %e %j";
+      }
+    }
+    ```
+
+  `snare` informs users whose config contains `repodir` how to update it.
+
+
+## Minor changes
+
+* `snare` now validates input derived from the webhook request so that it is
+  safe to pass to the shell: GitHub owners, repositories, and events are all
+  guaranteed to satisfy the regular expression `[a-zA-Z0-9._-]+` and not to be
+  the strings `.` or `..`.
+
+
 # snare 0.1.0 (2020-02-13)
 
 First release.
--- a/README.md
+++ b/README.md
@@ -43,8 +43,8 @@ The minimal recommended configuration file is:
 listen = "<ip-address>:<port>";
 
 github {
-  reposdir = "<path>";
   match ".*" {
+    cmd = "/path/to/prps/%o/%r %e %j";
     email = "<email-address>";
     secret = "<secret>";
   }
@@ -53,19 +53,20 @@ github {
 
 where:
 
-  * `ip-address` is either an IPv4 or IPv6 address and `port` a port on which an
-    HTTP server will listen.
-  * `path` is the directory where the per-repo programs are stored. For a
-    repository `repo` owned by `user` the command:
+ * `ip-address` is either an IPv4 or IPv6 address and `port` a port on which an
+   HTTP server will listen.
+ * `cmd` is the command that will be executed when a webhook is received. In
+   this case, `/path/to/prps` is a path to a directory where per-repo programs
+   are stored. For a repository `repo` owned by `user` the command:
 
-      ```
-      <reposdir>/<user>/<repo> <event> <path to GitHub JSON>
-      ```
+     ```
+     /path/to/prps/<user>/<repo> <event> <path-to-github-json>
+     ```
 
-    will be run. The file `<repo>` must be executable. Note that per-repo
-    programs are run with their current working directory set to a temporary
-    directory to which they can freely write and which will be automatically
-    removed when they have completed.
+   will be run. The file `<repo>` must be executable. Note that commands are
+   run with their current working directory set to a temporary directory to
+   which they can freely write and which will be automatically removed when
+   they have completed.
  * `email-address` is an email address to which any errors running per-repo
    programs will be sent (warning: full stderr/stdout will be sent, so consider
    carefully whether these have sensitive information or not). This uses
@@ -80,14 +81,15 @@ snare.conf](https://softdevteam.github.io/snare/snare.conf.5.html) contains the
 complete list of configuration options.
 
 
-## Per-repo programs
+## Commands
 
-When using snare, the *per-repo programs* do the actual work of executing
-specific actions for a given repository.  For example, `snare`'s GitHub
+`snare` can be used to run any command runnable from the Unix shell. The
+"per-repo program" model as documented above is one common way of doing this.
+For example, `snare`'s GitHub
 repository is
 [`https://github.com/softdevteam/snare`](https://github.com/softdevteam/snare).
 If we set up a web hook up for that repository that notifies us of pull request
-events, then the command:
+events, then with the above `snare.conf`, the command:
 
 ```sh
 <repo-programs-dir>/softdevteam/snare pull_request /path/to/json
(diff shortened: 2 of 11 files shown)


CHANGES.md
```diff
@@ -1,3 +1,39 @@
...
+
+    ```
+    github {
+      match ".*" {
+        cmd = "/path/to/reposdir/%o/%r %e %j";
+      }
+    }
+    ```
+
+  in order to obtain the previous behaviour. `snare` informs users whose config
```
vext01: I'd remove "in order to obtain the previous behaviour". I think that's obvious.

ltratt: Agreed. Fixed in https://github.com/softdevteam/snare/pull/43/commits/a78d6dc9e4a200132d4f1a6cfa736a9226ffefbb.

vext01: > `snare` informs users whose config contains `repodir` how to update it.

The grammar here feels somehow clumsy.

How about:

> Upon encountering `repodir`, `snare` will now print a message to help the user to migrate.

ltratt: I must admit that I don't like the current phrasing very much, but I think making it long-winded makes it even more clumsy :/ I'm not sure it's worth wasting too much time over TBH.

README.md
```diff
@@ -43,8 +43,8 @@ The minimal recommended configuration file is:
 listen = "<ip-address>:<port>";
 
 github {
-  reposdir = "<path>";
   match ".*" {
+    cmd = "/path/to/prps/%o/%r %e %j";
```
vext01: What is `prps` here? Per-repo programs? I think we ditched that terminology a while ago didn't we?

ltratt: We didn't ditch the terminology, but I'm open to suggestions of alternative names, if it makes things clearer. Note that the README does now say something along the lines of "one way of organising things is as per-repo programs."

README.md
```diff
@@ -53,19 +53,20 @@ github {
 
 where:
 
-  * `ip-address` is either an IPv4 or IPv6 address and `port` a port on which an
```
vext01: Here the unrelated ip-address part changed due to an indent change. Was that intentional?

ltratt: Sort of. I found an indentation bug in the README while reformatting. I should probably have made a separate commit.

README.md
```diff
@@ -53,19 +53,20 @@ github {
...
+   HTTP server will listen.
+ * `cmd` is the command that will be executed when a webhook is received. In
+   this case, `/path/to/prps` is a path to a directory where per-repo programs
+   are stored. For a repository `repo` owned by `user` the command:
 
-      ```
-      <reposdir>/<user>/<repo> <event> <path to GitHub JSON>
-      ```
+     ```
+     /path/to/prps/<user>/<repo> <event> <path to GitHub JSON>
```
vext01: We should probably use `<path-to-github-json>` so be consistent with other notations in this repo.

ltratt: Agreed. Fixed in https://github.com/softdevteam/snare/pull/43/commits/535a30bba7493feeef32ea47f6797bdbe829fc0b.

README.md
vext01: I think i might be wrong here, as i've since seen lots of references to that term.

I thought `reposdir` had replaced `per-repo-programs`. Are we mixing terminology to freely perhaps?

ltratt: Originally, a `reposdir` would contain "per repo programs", and that structure was enforced by snare-0.1.0. This PR removes the enforcement of either concepts, but that doesn't mean that the terminology isn't vaguely useful, since I think most people (or, at least us), will structure things as before.

src/config.rs
```diff
@@ -292,18 +277,35 @@ impl GitHub {
...
+    /// Verify that the `cmd` string is valid, returning `Ok())` if so or `Err(String)` if not.
+    pub(crate) fn verify_cmd_str(cmd: &str) -> Result<(), String> {
+        let mut i = 0;
+        while i < cmd.len() {
+            if cmd[i..].starts_with("%") {
+                if i + 1 == cmd.len() {
+                    return Err("'cmd' cannot end with a single '%'.".to_owned());
+                }
+                let c = cmd[i + 1..].chars().nth(0).unwrap();
+                if !(c == 'e' || c == 'o' || c == 'r' || c == 'j' || c == '%') {
```
vext01: What if you need a literal `%e` (for example) in the command which you don't want interpreted by snare?

ltratt: `%%e`

vext01: `date +%e` would make a nice test case.

ltratt: `date +%%e`

src/httpserver.rs
```diff
@@ -109,29 +109,14 @@ async fn handle(req: Request<Body>, snare: Arc<Snare>) -> Result<Response<Body>,
...
-            // we (and the OS) are probably in deep trouble anyway...
-            nix::unistd::write(snare.event_write_fd, &[0]).ok();
-            return Ok(res);
-        }
-    }
-
-    snare.error(&format!("No per-repo program {}/{}.", owner, repo));
-    // We couldn't find a per-repo program for this request.
-    *res.status_mut() = StatusCode::BAD_REQUEST;
+    let repo_id = format!("github/{}/{}", owner, repo);
```
vext01: I wonder if we should push to a `PathBuf` here instead of construct path separators manually? Since we don't expect this to work on !unix it probably doesn't really matter though.

ltratt: This is intentionally not a file system path: it's more like a URL. It's only used to identify one repository vs. another in the queue. At the moment the `github/` prefix is entirely redundant, but it's there in case we support e.g. gitlab in the future.

src/jobrunner.rs
```diff
@@ -487,9 +512,60 @@ impl JobRunner {
...
+            } else if c == 'o' {
+                cmd.push_str(owner);
+                i += 2;
+            } else if c == 'r' {
+                cmd.push_str(repo);
+                i += 2;
+            } else if c == 'j' {
+                cmd.push_str(json_path);
+                i += 2;
+            } else if c == '%' {
```
vext01: Ah, so this answers one of my earlier questions.

CHANGES.md
vext01: Let me distill out what I dislike about the sentence. Eliding the middle part of the sentence, we have:

> snare informs users how to update it

Sounds a bit cave man.

> > snare informs users **on** how to update it

Seems to make it better. So does this work perhaps?
> snare informs users whose config contains repodir **on** how to update it.

If not, agreed, let's move on.

README.md
vext01: (sorry, we raced, so my last comment seems out of context).

So are is the `per-repo-programs` directory different from the `reposdir`? I think so. In which case we need separate terms anyway.

README.md
vext01: That's ok. Just checking.

src/httpserver.rs
vext01: OK, I was totally wrong here :)

src/config.rs
vext01: Yep, agreed. Missed this in the manual first time around.

CHANGES.md
ltratt: `on` doesn't feel grammatically correct either :( While I agree the current phrasing sucks, I think we should move on.

README.md
ltratt: Yes.

src/httpserver.rs
ltratt: :)

CHANGES.md
vext01: yep

vext01:
OK, that's all I've got. Please squash.

ltratt:
Squashed.

PR #44, opened 2020-02-25
String escapes

ltratt:
Process escaped strings in the config file.
    
The regex in config.l supported strings such as:

```    
cmd = "\"\\";
```

but we didn't actually process the `\` characters, which was something of an oversight!

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -73,9 +73,8 @@ impl Config {
                                     "Mustn't specify 'listen' more than once",
                                 ));
                             }
-                            let listen_str = lexer.span_str(span);
-                            let listen_str = &listen_str[1..listen_str.len() - 1];
-                            match SocketAddr::from_str(listen_str) {
+                            let listen_str = unescape_str(lexer.span_str(span));
+                            match SocketAddr::from_str(&listen_str) {
                                 Ok(l) => listen = Some(l),
                                 Err(e) => {
                                     return Err(error_at_span(
@@ -127,9 +126,8 @@ impl Config {
                                     "Mustn't specify 'user' more than once",
                                 ));
                             }
-                            let user_str = lexer.span_str(span);
-                            let user_str = &user_str[1..user_str.len() - 1];
-                            user = Some(user_str.to_owned());
+                            let user_str = unescape_str(lexer.span_str(span));
+                            user = Some(user_str);
                         }
                     }
                 }
@@ -172,8 +170,7 @@ impl GitHub {
         }
 
         for m in ast_matches {
-            let re_str = lexer.span_str(m.re);
-            let re_str = format!("^{}$", &re_str[1..re_str.len() - 1]);
+            let re_str = format!("^{}$", unescape_str(lexer.span_str(m.re)));
             let re = match Regex::new(&re_str) {
                 Ok(re) => re,
                 Err(e) => {
@@ -199,10 +196,9 @@ impl GitHub {
                                 "Mustn't specify 'cmd' more than once",
                             ));
                         }
-                        let cmd_str = lexer.span_str(span);
-                        let cmd_str = &cmd_str[1..cmd_str.len() - 1];
-                        GitHub::verify_cmd_str(cmd_str)?;
-                        cmd = Some(cmd_str.to_owned());
+                        let cmd_str = unescape_str(lexer.span_str(span));
+                        GitHub::verify_cmd_str(&cmd_str)?;
+                        cmd = Some(cmd_str);
                     }
                     config_ast::PerRepoOption::Email(span) => {
                         if email.is_some() {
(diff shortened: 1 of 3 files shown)


src/config.rs
```diff
@@ -353,6 +347,34 @@ impl GitHub {
     }
 }
 
+/// Take a quoted string from the config file and unescape it (i.e. strip the start and end quote
+/// (") characters and process any escape characters in the string.)
+fn unescape_str(us: &str) -> String {
+    // The regex in config.l should have guaranteed that strings start and finish with a
+    // single quote (") character.
```
vext01: A single quote. Shows a double quote in brackets. Hilarity ensues :)

src/config.rs
```diff
@@ -353,6 +347,34 @@ impl GitHub {
...
+    debug_assert!(us.starts_with('"') && us.ends_with('"'));
+    let mut s = String::new();
+    // We iterate over all characters except the opening and closing quote characters.
+    let mut i = '"'.len_utf8();
```
vext01: Shall we cache the length of a double quote in a local variable, as we compute it lots of times? I'd also be OK with asserting the length is 1 and using the constant 1 henceforth.

src/config.rs
```diff
@@ -353,6 +347,34 @@ impl GitHub {
...
+    while i < us.len() - '"'.len_utf8() {
+        let c = us[i..].chars().nth(0).unwrap();
```
vext01: Do I see it right that `nth(0)` is the same as `next()` here?

src/config.rs
ltratt: They'll almost certainly be turned into a constant by the compiler, and it makes sure we can't forget that characters aren't always 1 byte wide.

src/config.rs
ltratt: Huh?

src/config.rs
ltratt: Yes I guess they are. I don't really prefer one over the other.

src/config.rs
vext01: > The regex in config.l should have guaranteed that strings start and finish with a single quote (") character.

But there's a double quote in the brackets.

I suspect you meant to say double quote?

ltratt: Oh, I see! It means "single double quote" if that makes sense?

src/config.rs
vext01: That rebuttals the second suggestion.

So how about replacing all of the instances of `'"'.len_utf8()` (which is qite hard to read due to syntactical noise) with `dquo_len` where `dquo_len = '"'.len_utf8()`?

src/config.rs
vext01: I see. I never would have gotten that from the prose. "single double quote" also offers the reader the suspicion that there may be a typo.

I think we should be really captain obvious about it and say:
> ...should have guaranteed that strings start and finish with a double quote (") character.

src/config.rs
ltratt: Both ways suck, but I find the overhead of a variable even higher than the repeated function call.

src/config.rs
ltratt: I agree: the phrasing was pretty terrible! Fixed in https://github.com/softdevteam/snare/pull/44/commits/7c6d4032bc4c3ae415d5f57099fc6c8d3b7e8ace.

src/config.rs
vext01: Personally I prefer to have the easily-read variable name. I'd have thought the Rust compiler would inline it to a constant anyway.

I won't insist. Do whatever you think is best and we can move on.

ltratt: When I said "I find the overhead higher" I meant the "intellectual overhead" (the compiler will optimise both cases away).

src/config.rs
vext01: I think next is about 1 byte more succinct :)

Again, I won't get hung up on this.

ltratt: I must admit, I wish there was a way that didn't require an `unwrap`. I have a marginal preference for `nth(0)` simply because it looks a bit more like `[0]` which is what we'd really like to write.

ltratt: It turns out that if you do `"blah"[0]` the compiler suggests that you use `chars().nth(0)` so I think we might as well stick with what we've got.

[I understand why they removed `char_at`, but I wish they'd kept it in!]

ltratt:
I think this is ready for squashing?

vext01:
Please squash

ltratt:
Squashed.

vext01:
Hrm:
> error[E0425]: cannot find function `unescape_string` in this scope

ltratt:
That error is entirely my idiocy. Fixed in https://github.com/softdevteam/snare/pull/44/commits/6355c8fd53605de0ef2a458327a163c0effcf39b.

vext01:
Please squash

ltratt:
Squashed.

And at least we know that `cargo test` works :p

PR #45, opened 2020-02-25
Turn lexing errors into parsing errors.

ltratt:
Consider this (incorrect) snippet of a snare.conf:

```
  listen = "0.0.0.0:8011";
  maxjob = 4;
```

Before this commit you would simply be told:

```
  Lexing error at line 2 column 1.
```

which is accurate, but not very helpful. This commit adds a simple rule which, if no other lexing rules apply, consumes one character of the input into an "UNKNOWN" token. Thus every possible input can now be lexed, and lexing errors are now replaced by parsing errors, which means that error recovery can now kick in. Thus the above snippet now leads to:

```
  Parsing error at line 2 column 1. Repair sequences found:
     1: Insert MAXJOBS, Delete m, Delete a, Delete x, Delete j, Delete o, Delete b.
```

### DIFF
--- a/src/config.l
+++ b/src/config.l
@@ -20,4 +20,5 @@ sequential "SEQUENTIAL"
 timeout "TIMEOUT"
 user "USER"
 //.*?$ ;
-[ \t\n\r]* ;
+[ \t\n\r]+ ;
+. "UNKNOWN"
--- a/src/config.y
+++ b/src/config.y
@@ -67,6 +67,11 @@ QueueKind -> Result<(Span, QueueKind), ()>:
   | "SEQUENTIAL" { Ok((map_err($1)?, QueueKind::Sequential)) }
   ;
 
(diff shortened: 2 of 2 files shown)


ptersilie:
This looks fine. Though I was surprised to see that the syntax checking happens in the lexer in the first place. I was expecting the config grammar to be something like `ID = value`, where ID matches `[a-z]+` and we check during/after parsing if `ID` was a valid config option.

ltratt:
Because the config only has a small number of options, you get, IMHO, better error messages if it says "Insert MA-JOBS" rather than "Insert ID" and so on. It's a matter of taste though.

ptersilie:
I've just looked at the grammar and can see now that by using action code you could defer a lot of the checking into the parser, which probably saves a lot of post-parsing code. I was still in "we-don't have action code" mode.

bors r+

PR #53, opened 2020-03-06
Fix config file

ltratt:


snare.conf.example
```diff
@@ -5,10 +5,13 @@
 listen = "<address>:<port>";
 
 github {
-  // The path to the directory where your per-repo programs are stored.
-  reposdir = "<path>";
-
   match ".*" {
+    // For each webhook coming in, this will execute the command:
```
vext01: I wonder, can we add a test which checks the example config file is valid perhaps?

vext01:
Just one small comment

snare.conf.example
ltratt: I had to tweak things a bit but this is a good idea: fixed in https://github.com/softdevteam/snare/pull/53/commits/13d9afbe307501a0bd43bcebbbd6e06fc16c77d5.

vext01:
LGTM. Please squash.

ltratt:
No need to squash IMHO.

PR #54, opened 2020-03-06
Don't search for a per-user snare.conf.

ltratt:
And assume the default location is /etc/snare/snare.conf, which gives a bit of flexibility if, in the future, we want to add other files to snare.


vext01:
LGTM

bors r+

PR #55, opened 2020-03-06
Always chdir to '/'.

ltratt:
Previously we left snare with its CWD unchanged unless a `user` was specified at
which point we (for no obvious reason) changed CWD to `user`'s `$HOME`. This
small difference seems pointless: by always setting CWD to `/` we lessen the
chance of users relying on the CWD.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -188,18 +188,6 @@ fn change_user(conf: &Config) {
     match conf.user {
         Some(ref user) => match get_user_by_name(&user) {
             Some(u) => {
-                if let Err(e) = set_current_dir(u.home_dir()) {
-                    fatal_err(
-                        &format!(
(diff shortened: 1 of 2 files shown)


src/main.rs
```diff
@@ -259,11 +247,13 @@ pub fn main() {
 
     change_user(&conf);
 
+    set_current_dir("/").unwrap_or_else(|_| fatal("Can't chdir to '/'"));
     if daemonise {
         if let Err(e) = daemon(true, false) {
             fatal_err("Couldn't daemonise: {}", e);
         }
     }
+
```
vext01: intentional?

ltratt: In this case, yes :)

vext01:
I was going to say: let's switch to `/var/empty`, but linux doesn't have it :\

So I'm happy with this.

bors r+

PR #61, opened 2020-03-13
Fix and slightly extend the example email differ.

ltratt:
The previous version did not define the `$repo_url` variable. This version does, and also supports emailing multiple people (which is a small, but useful, extension).


vext01:
LGTM.

bors r+

PR #63, opened 2020-03-14
Ensure that no errors are sent to stderr after daemonisation.

ltratt:
The main commit here is https://github.com/softdevteam/snare/commit/6f452401ef187f6d34f9351670566e3ff069a919, but there are two warm-up commits, and one I-noticed-it-while-I'm-here commit (https://github.com/softdevteam/snare/commit/cb7dd47c851ecc522649f327761ec18d30cae7cf) to round things off.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -17,7 +17,6 @@ use std::{
     error::Error,
     ffi::CString,
     fmt::Display,
-    io::{stderr, Write},
     os::unix::io::RawFd,
     path::{Path, PathBuf},
     process,
@@ -202,7 +201,7 @@ fn usage(prog: &str) -> ! {
         .file_name()
         .map(|x| x.to_str().unwrap_or("snare"))
         .unwrap_or("snare");
-    writeln!(&mut stderr(), "Usage: {} [-c <config-path>] [-d]", leaf).ok();
+    eprintln!("Usage: {} [-c <config-path>] [-d]", leaf);
     process::exit(1)
 }
 
--- a/src/main.rs
+++ b/src/main.rs
@@ -143,22 +143,6 @@ impl Snare {
     }
 }
 
-/// Exit with a fatal error. This function should only be called before the [`Snare`](::Snare)
-/// struct is created.
-fn fatal(msg: &str) -> ! {
-    if msg.ends_with('.') {
-        eprintln!("{}", msg);
-    } else {
-        eprintln!("{}.", msg);
-    }
-    process::exit(1);
-}
-
-/// Exit with a fatal error, printing the contents of `err`.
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -117,19 +117,7 @@ impl Snare {
     ///
     /// If `msg` contains a `NUL` byte.
     fn fatal(&self, msg: &str) -> ! {
-        if self.daemonised {
```
vext01: So if I understand correctly, this removed code worked, but was unnecessary, as we can just delegate the logic to `fatal()`?

src/main.rs
```diff
@@ -153,48 +141,58 @@ fn search_snare_conf() -> Option<PathBuf> {
 }
 
 /// If the config specified a 'user' then switch to that and update $HOME and $USER appropriately.
+/// This function *must* not be called after daemonisation.
```
vext01: Ah. I think you should emphasize *not* instead of *must*. It's very easy to see:
> This function *must* be called after daemonisation.

src/main.rs
ltratt: Previously `fatal` wasn't powerful enough, but because it now takes a `daemonised` argument, it can subsume this call.

src/main.rs
ltratt: To be honest, the emphasis is overkill. Removed in https://github.com/softdevteam/snare/pull/63/commits/7ae0c1b2395c9acd4bca66adb5b56f2fd528b051.

vext01: OK. There is at least one other instance of emphasis added in this PR.

ltratt: Removed in https://github.com/softdevteam/snare/pull/63/commits/4faf5eea066d3647b5b034509310d66250078eae.

src/main.rs
```diff
@@ -153,48 +141,58 @@ fn search_snare_conf() -> Option<PathBuf> {
...
     }
 }
 
-/// Exit with a fatal error. This function should only be called before the [`Snare`](::Snare)
-/// struct is created.
-fn fatal(msg: &str) -> ! {
-    if msg.ends_with('.') {
-        eprintln!("{}", msg);
+/// Exit with a fatal error.
+fn fatal(daemonised: bool, msg: &str) -> ! {
```
vext01: I just looked to see if there's a way to know if we are daemonisaed so that we don't have to pass this flag around. That's a can of worms!

The best option I found (for us) was to check if stderr is open:
https://cboard.cprogramming.com/networking-device-communication/93683-checking-if-file-descriptor-valid.html#post678998

But it seems hardly worth the effort.

ltratt: Indeed, I looked into that, and came to the same conclusion.

vext01:
Please squash.

ltratt:
Squashed.

PR #64, opened 2020-03-16
Remove direct mio dependency

goffrie:
This seems vestigial.


ltratt:
You're quite right -- good spot!

bors r+

PR #65, opened 2020-03-20
Deprecate the `email` option and replace it with `errorcmd`.

ltratt:
The `email` option always felt a bit over-specific; this PR generalises `email` to `errorcmd` in the same way that 700ab030 generalised `reposdir` to `cmd`. `errorcmd` and `cmd` deliberately share almost identical syntax (the only difference being that `errorcmd` accepts an additional `%s` over `cmd`).

To obtain the previous behaviour:

```
email = "someone@example.com";
```

should be changed to:

```
errorcmd = "cat %s | mailx -s \"snare error: github.com/%o/%r\" someone@example.com";
```

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,29 @@
+# snare 0.4.0 (2020-xx-xx)
+
+## Breaking changes
+
+* The `email` option in `match` blocks has been replaced by the more generic
+  `errorcmd`. To obtain the previous behaviour:
+
+    ```
+    email = "someone@example.com";
+    ```
+
+  should be changed to something like:
+
+    ```
+    errorcmd = "cat %s | mailx -s \"snare error: github.com/%o/%r\" someone@example.com";
+    ```
+
+  This assumes that the `mailx` command is installed on your machine.  As this
+  example may suggest, `errorcmd` is much more flexible than `email`.  The
+  syntax of `errorcmd` is the same as `cmd` with the addition that `%s` is
+  expanded to the path of the failed job's combined stderr / stdout.
+
+  `snare` informs users whose config contains `email` how to update to
+  `errorcmd` to obtain the previous behaviour.
+
+
 # snare 0.3.0 (2020-03-08)
 
 ## Breaking changes
--- a/snare.conf.5
+++ b/snare.conf.5
@@ -129,12 +129,40 @@ or
 .Qq .. .
 This means that they are safe to pass as shell arguments and/or to be included
 in file system paths.
-.It Sy email = Qq Em address ;
-optionally specifies an email address to which any errors when running commands
-will be sent (warning: full stderr/stdout will be sent, so consider carefully
-whether these have sensitive information or not).
-This uses the `sendmail` command to send email: you should ensure that you have
-installed, set-up, and enabled a suitable `sendmail` clone.
+.It Sy errorcmd = Qq Em shell-cmd ;
+optionally specifies a command to be run when a job exits unsuccessfully.
+.Em shell-cmd
+will be executed via
+.Ql $SHELL -c .
+The following escape sequences are recognised and replaced before execution:
+.Bl -tag -width Ds
+.It Sy %e
+the GitHub event type (e.g.
+.Ql pull_request ) .
+.It Sy %j
+the path to the GitHub JSON.
+.It Sy %o
+the repository owner.
+.It Sy %r
+the repository.
+.It Sy %s
+the path to the file containing the job's combined stderr / stdout.
+.It Sy %%
+a literal
+.Ql % .
+.El
+Note that
+.Ql %
+may not be followed by any character other than those above.
+.Pp
+The escape sequences are guaranteed to satisfy the regular expression
+.Qq [a-zA-Z0-9._-]+
+and not to be the strings
+.Qq \&.
+or
+.Qq .. .
+This means that they are safe to pass as shell arguments and/or to be included
+in file system paths.
 .It Sy queue = Po evict | parallel | sequential Pc ;
 specifies what to do when multiple requests for the same repository
 are queued at once:
@@ -202,7 +230,7 @@ listen = "<address>:<port>";
 github {
   match ".*" {
     cmd = "/path/to/prps/%o/%r %e %j";
-    email = "<email>";
+    error_cmd = "cat %s | mailx -s \\"snare error: github.com/%o/%r\\" someone@example.com";
     secret = "<secret>";
   }
 }
@@ -215,6 +243,9 @@ Each repository then has a unique program
 .Qq %o/%r
 which will be executed with two arguments: the GitHub event; and the path to
 the GitHub JSON.
+If a job exits unsuccessfully then an email will be sent to someone@example.com
+containing the job's comined stderr and stdout output (assuming that a suitable
+sendmail clone has been installed and activated).
 .Pp
 The top-to-bottom evaluation of match blocks allow users to specify defaults
 which are only overridden for specific repositories.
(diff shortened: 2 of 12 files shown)


CHANGES.md
```diff
@@ -1,3 +1,28 @@
...
+
+  should be changed to:
+
+    ```
+    errorcmd = "cat %s | mailx -s \"snare error: github.com/%o/%r\" someone@example.com";
+    ```
+
+  As this example may suggest, `errorcmd` is much more flexible than `email`.
+  The syntax of `errorcmd` is the same as `cmd` with the addition that `%s` is
+  expanded to the path of the failed job's combined stderr / stdout.
```
vext01: This is what I was confused about before.

So since snare writes stdout/err to disk, you would never want your `errorcmd` to take input from a pipe anyway (unless you changed snare).

CHANGES.md
```diff
@@ -1,3 +1,28 @@
...
+## Breaking changes
+
+* The `email` option in `match` blocks has been replaced by the more generic
+  `errorcmd`. To obtain the previous behaviour:
+
+    ```
+    email = "someone@example.com";
+    ```
+
+  should be changed to:
```
vext01: should -> could?

There's any number of mail clients you could use, or I suppose you could even invoke sendmail directly.

CHANGES.md
```diff
@@ -1,3 +1,28 @@
+# snare 0.4.0 (2020-xx-xx)
+
+## Breaking changes
+
+* The `email` option in `match` blocks has been replaced by the more generic
+  `errorcmd`. To obtain the previous behaviour:
```
vext01: Did you want to mention the motivation? I think it's so that you can notify by other means if you so wish, right? Like slack / mattermost / irc / ...

snare.conf.5
```diff
@@ -225,11 +256,11 @@ github {
   reposdir = "<path>";
   match ".*" {
     cmd = "/path/to/prps/%o/%r %e %j";
-    email = "abc@def.com";
+    error_cmd = "cat %s | mailx -s \\"snare error: github.com/%o/%r\\" abc@def.com";
     secret = "sec";
   }
   match "a/b" {
-    email = "ghi@jkl.com";
+    error_cmd = "xyz";
```
vext01: Shall we make that a real command to demonstrate another use of errorcmd?

src/jobrunner.rs
```diff
@@ -479,34 +479,37 @@ impl JobRunner {
...
+                &job.json_path.as_os_str().to_str().unwrap(),
+                &job.stderrout.path().as_os_str().to_str().unwrap(),
+            );
+            let mut child = match Command::new(&self.shell)
+                .arg("-c")
+                .arg(&errorcmd)
+                .current_dir(job.tempdir.path())
+                .stderr(process::Stdio::null())
+                .stdout(process::Stdio::null())
+                .stdin(process::Stdio::null())
```
vext01: Did we want to do anything with the output of a failing errorcmd? Not worth the effort?

vext01:
Just a few small comments.

CHANGES.md
ltratt: I don't understand what you mean.

CHANGES.md
ltratt: I think "could" is just unnecessarily vague. If someone realises there's a better way for them, they'll find it. Everyone else can copy and paste this almost unthinkingly and be all-but-sure it'll work.

snare.conf.5
ltratt: I would love to, but I failed to think of a good example :( Suggestions are very welcome!

src/jobrunner.rs
ltratt: How many levels deep do you want to go? :) If the `errorcmd` fails then we report that failure to the log, but I don't think it would be appropriate to send an unbounded amount of stderr/stdout there.

CHANGES.md
ltratt: Is "more generic" not sufficient motivation? I'm happy to consider something else if you think not.

CHANGES.md
vext01: In the snare_scripts repo I was questioning where the file came from, and this explains.

No action necessary on your part.

ltratt: OK.

CHANGES.md
vext01: > Everyone else can copy and paste this almost unthinkingly and be all-but-sure it'll work.

I disagree. `mailx` is not standard. For example, I have a fresh Debian install here, and it has neither `mailx` nor `mail`.

CHANGES.md
vext01: Yeah OK.

CHANGES.md
ltratt: Every box I've seen that has an active sendmail has `mailx`. So if the `email` options worked before, I'm relatively confident that `mailx` inside `errorcmd` will work now.

snare.conf.5
vext01: Most of the actually useful ones are going to be quite involved, so how about something fun?

```
lp %s
```

```
notify-send "It's broken Dave" "`cat %s`"
```

src/jobrunner.rs
vext01: I buy your reasoning. OK.

snare.conf.5
ltratt: What's `lp`? `notify-send` I considered, but it seemed unlikely that someone running a daemon like snare would do so on a desktop!

vext01: What's `lp`? That's how you used to print things to your dot-matrix printer back in 1982 :)

> notify-send I considered, but it seemed unlikely that someone running a daemon like snare would do so on a desktop

Yeah, but it's just for fun! A command that a reader recognises must be better than `xyz`.

If you wanted something more serious, you could append the error to a file some place.

ltratt: Well, we could do `echo "Error in %o/%r" | lpr -` which I think will work on a modern machine?

CHANGES.md
vext01: OK, so I just installed sendmail on the Debian box. Still no `mailx` or `mail` :P

Really, I don't think we should encourage copy-pasting.

I'm not going to die on this hill, so if you still disagree, I'll back down.

snare.conf.5
vext01: `lpr` is fine.

Wouldn't you need a `%s` though? I thought snare put stderr/stdout to file?

ltratt: You don't *have* to use any of the `%` modifiers if you don't want to. But yeah `lrp %s` has the virtue of simplicity!

CHANGES.md
ltratt: OK, how about I expand it to "if you have `mailx` installed"?

vext01: Sure.

snare.conf.5
vext01: I'm missing some detail about stdout/stderr.

So if `%s` is absent, it pushes the error text to stdin of the errorcmd?

ltratt: Fixed in https://github.com/softdevteam/snare/pull/65/commits/ffa53cf2c1095d001bb82b8d69b6aa0e3c7b90df.

CHANGES.md
ltratt: Fixed in https://github.com/softdevteam/snare/pull/65/commits/54c6ad864a2809931c57000c8e5c3ac7c2b53910.

snare.conf.5
ltratt: > I'm missing some detail about stdout/stderr.

Have you read the man page? Does that explain it adequately?

vext01: I misread.

You are proposing: `echo "Error in %o/%r" | lpr -`, which doesn't include the error text at all.

So OK.

vext01:
Please squash

ltratt:
Squashed.

PR #66, opened 2020-03-21
Use SIGCHLD to listen for child process exit rather than using a timeout

ltratt:
This saves snare from having to wake-up unnecessarily often.


vext01:
Looks like bors crashed, so I will retry.

bors r+

PR #76, opened 2020-08-06
Validate git commit hashes before passing them to `git log`.

ltratt:
Before this, if GitHub went bad, they could possibly pass something that would cause `git log` to (at least) leak more information than we expected.


vext01:
Do we need a comment saying what the line does? It may not be obvious for new users.

ltratt:
I agree that many people might not realise what it's doing, but when I tried explaining it, I felt it was a bit awkward. I ended up thinking that the paranoid will realise what's going on and the non-paranoid will just overlook it...

vext01:
OK.

bors r+

PR #80, opened 2020-12-02
Update dependencies

ltratt:
The first commit is just `cargo update` mostly so that we get a `nix` that works on OpenBSD. The second commit does a could-have-been-more-significant-but-actually-isn't upgrade of `signal-hook`. Tested working on my server.


vext01:
Fine.

bors r+

PR #83, opened 2021-02-15
Move to tokio 1.

ltratt:
The last time I tried this, it failed, as some dependencies still relied on tokio < 1, and we ended up with two tokio runtimes being created, with the obvious outcome that nothing worked. It now seems that all the dependencies are
now working with tokio >= 1, so we can finally move forwards. I've also had to do "cargo update", so there's lots of obfuscating churn in `Cargo.lock`: the real changes are in `Cargo.toml`.

I've lightly tested this on b13 without issue and I'll keep it running on there for a while to see if anything crops up. However, I don't think there's a good reason not to consider merging this now.

### DIFF
--- a/.buildbot.sh
+++ b/.buildbot.sh
@@ -17,5 +17,9 @@ mkdir test_install
 PREFIX=test_install make install
 test -f test_install/bin/snare
 
-which cargo-deny | cargo install cargo-deny
-cargo-deny check license
+which cargo-deny | cargo install cargo-deny || true
+if [ "X`which cargo-deny`" != "X"]; then
+    cargo-deny check license
+else
+    echo "Warning: couldn't run cargo-deny" > /dev/stderr
+fi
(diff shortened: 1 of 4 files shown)


vext01:
LGTM, and there are now zero security problems reported.

Would you like me to merge this then?

ltratt:
I don't see any harm in doing so. *If* something crops up over a few weeks of testing, I can always do a revert, though I'm not really expecting to need to do so.

vext01:
Oops. Something has gone quite wrong!

ltratt:
It looks like `cargo deny` isn't building. I'm investigating.

ltratt:
bors try

ltratt:
bors try

ltratt:
OK for me to squash?

vext01:
Aha, Dr. `set -e`, to what do I owe this pleasure? 😆 

Please squash.

ltratt:
Squashed. And I'll remember `set -e` next time... maybe ;)

PR #102, opened 2022-02-09
Snare 0.4.5

ltratt:


ltratt:
> Is it worth bringing the Cargo.toml deps in line with Cargo.lock? I'm uncertain.

No, because we ship a `Cargo.lock`, that serves no purposes. [`Cargo.toml` specifies minimum versions and `Cargo.lock` is free to use later versions iff they're compatible.]

vext01:
Indeed, that was why i was unsure.

bors r+

PR #109, opened 2023-01-12
Update dependencies

ltratt:
It's quite a while since we updated snare's dependencies, so this PR does it slowly in stages. This has already survived light testing; if it survives for a few days running in anger, I'll probably make a new release, but that'll be a separate PR.


vext01:
Any squashing to do?

ltratt:
I think they're best kept separate in this unusual instance.

vext01:
As you wish (although no update required any actual code change, so I'd have done this all in one go using `cargo upgrade`).

bors r+

PR #112, opened 2023-01-23
Move from `json` to `serde_json`.

ltratt:
After updating a couple of other dependencies, this PR minimally moves from the unmaintained `json` crate to `serde_json`.


vext01:
Thanks for fixing. Looks fairly painless.

I take it there is no squashing to do?

ltratt:
No squashing needed.

PR #117, opened 2023-11-07
Migrate to github merge queues.

vext01:


vext01:
Failed off the bat :P Investigating.

vext01:
Investigating how to re-run the check. Hoping we don't have to close and re-open. That would be annoying for (e.g.) transient errors.

ltratt:
> Instead I logged in to the buildbot web ui and clicked "rebuild".

That seems to have updated the status check, so that's actually not *so* bad.

vext01:
> That seems to have updated the status check

Yes, I needed that, or the PR would never be deemed ready.

vext01:
There we go!

PR #122, opened 2023-12-17
Remove hyper & tokio; implement an HTTP server; and add system tests.

ltratt:
This commit is a thorough yak shave. First I tried to upgrade hyper, which we previously used for snare's HTTP server. However, it has thoroughly changed its API and no longer provides an HTTP server directly (there is an external crate which is supposed to work, but I couldn't get it to compile). That slowly made me rethink snare's HTTP server, because the use of hyper is why snare also required tokio, with all that comes with it (async, lots of dependencies, and the overloading of my pea brain).

So, second, I removed hyper and tokio and implemented my own minimalistic HTTP server (heavily modelled on pizauth). This hugely reduces our dependencies (from 213 -> 172 for `cargo build`) and reduces the final binary size by 20% in release mode.

However, how do we know that taking on the burden of running our own HTTP server has left snare correct? One of snare's long-term weaknesses is that we haven't have good system-level tests, and I've often been lacking confidence when working on snare that my changes haven't broken something important.

So, third, and finally, this commit adds system level tests is `tests/basic.rs`. The most thorough of these fires a full GitHub-esque request at snare, forcing it to authenticate, and run a test command. While it's not perfect, it's a lot better than nothing.

This commit has also survived being run against GitHub itself. There is more to do (e.g. the tests currently bind to fixed ports; and there aren't enough system tests), but this feels like a sufficiently big unit of work in and of itself.

### DIFF
--- a/src/httpserver.rs
+++ b/src/httpserver.rs
@@ -1,175 +1,234 @@
-use std::{convert::Infallible, sync::Arc, time::Instant};
+use std::{
+    collections::HashMap,
+    error::Error,
+    io::{BufRead, BufReader, Read, Write},
+    net::{Shutdown, TcpListener, TcpStream},
+    sync::{
+        atomic::{AtomicUsize, Ordering},
+        Arc,
+    },
+    thread,
+    time::{Duration, Instant},
+};
 
 use hmac::{Hmac, Mac};
-use hyper::service::{make_service_fn, service_fn};
-use hyper::{self, body::Bytes, server::conn::AddrIncoming, Body, Request, Response, StatusCode};
 use percent_encoding::percent_decode;
 use secstr::SecStr;
-use sha1::Sha1;
+use sha2::Sha256;
 
 use crate::{queue::QueueJob, Snare};
 
-pub(crate) async fn serve(server: hyper::server::Builder<AddrIncoming>, snare: Arc<Snare>) {
-    let make_svc = make_service_fn(|_| {
-        let snare = Arc::clone(&snare);
-        async { Ok::<_, Infallible>(service_fn(move |req| handle(req, Arc::clone(&snare)))) }
-    });
+/// How many connections to accept simultaneously? Limiting this number stops attackers from
+/// causing us to use too many resources.
+static MAX_SIMULTANEOUS_CONNECTIONS: usize = 16;
+/// How long to try reading/writing from a socket before we assume it's died.
+static NET_TIMEOUT: Duration = Duration::from_secs(10);
+/// The maximum payload size we'll accept from a remote in bytes. The main reason to limit this is
+/// to stop large numbers of requests causing us to run out of memory.
+static MAX_HTTP_BODY_SIZE: usize = 64 * 1024;
 
-    if let Err(e) = server.serve(make_svc).await {
-        snare.fatal_err("Couldn't start HTTP server", e);
+pub(crate) fn serve(snare: Arc<Snare>) -> Result<(), Box<dyn Error>> {
+    let listener = TcpListener::bind(snare.conf.lock().unwrap().listen)?;
+    let active = Arc::new(AtomicUsize::new(0));
+    for mut stream in listener.incoming().flatten() {
+        // We want to keep a limit on how many threads are started concurrently, so that an
+        // attacker can't DOS the machine. `active` keeps track of how many threads are (or are
+        // just about to be) active. Since the common case is that we haven't hit the limit, we
+        // speculatively `fetch_add` and, if that fails, we then "undo" that with a `fetch_sub`,
+        // wait and try again. [Since the main thread is the only thread incrementing the count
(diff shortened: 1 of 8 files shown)


src/httpserver.rs
```diff
@@ -1,178 +1,228 @@
...
+static NET_TIMEOUT: Duration = Duration::from_secs(10);
+/// The maximum payload size we'll accept from a remote in bytes. The main reason to limit this is
+/// to stop large numbers of requests causing us to run out of memory.
+static MAX_HTTP_BODY_SIZE: usize = 64 * 1024;
+
+pub(crate) fn serve(snare: Arc<Snare>) -> Result<(), Box<dyn Error>> {
+    let listener = TcpListener::bind(snare.conf.lock().unwrap().listen)?;
+    let active = Arc::new(AtomicUsize::new(0));
+    for mut stream in listener.incoming().flatten() {
+        while active.fetch_add(1, Ordering::Relaxed) > MAX_SIMULTANEOUS_CONNECTIONS {
```
vext01, line 34: While adding one is greater than what we are permitting to accept, subtract one... This is clever in a way I'm not sure what it does, and needs a comment I reckon!

src/httpserver.rs
```diff
@@ -1,178 +1,228 @@
...
-        .get("X-Hub-Signature")
-        .and_then(|s| match s.to_str() {
-            Ok(s) => Some(s),
-            Err(_) => None,
-        })
-        .and_then(|s| s.split('=').nth(1))
-        .map(|s| s.to_owned())
+/// A very literal, and rather unforgiving, implementation of RFC2616 (HTTP/1.1), returning the URL
+/// of GET requests: returns `Err` for anything else.
+fn parse_get(stream: &mut TcpStream) -> Result<(HashMap<String, String>, Vec<u8>), Box<dyn Error>> {
```
vext01, line 147: This function is complicated and is bound to have bugs. It could benefit from unit testing.

Perhaps `TcpStream` is `Read` and this could be generalised to accepting `&str`s and `TcpStream`s, allowing easier testing?

src/httpserver.rs
ltratt: Hopefully addressed in 36f8f5e.

src/httpserver.rs
ltratt: It can't accept `&str`s because it has to do some interpretation (notably of `Content-Length`) before carrying on reading. [My first version didn't do that, but github's HTTP request library doesn't do "the right thing" if you try and simplify this!]

The good news is that the new system tests do partially test this. Not completely, of course, and I want to add more tests, but it is at least somewhat tested.

tests/basic.rs
```diff
@@ -0,0 +1,179 @@
...
+    )
+}
+
+#[test]
+fn bad_config() -> Result<(), Box<dyn Error>> {
+    run(r#""#, |_| Ok(()), exit_error)
+}
+
+#[test]
+fn full_request() -> Result<(), Box<dyn Error>> {
```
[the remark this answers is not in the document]
ltratt: Agreed. My aim is for the next PR to add several "negative" tests like this.

src/httpserver.rs
vext01: I think I get it.

src/httpserver.rs
vext01: I suppose in a way, by connecting to localhost and writing a string down a socket, you are somewhat accepting a string anyway.

I won't push, but it'd be good to test this function in isolation, if not now, later.

vext01:
Please squash.

src/httpserver.rs
ltratt: It's worth stating that the only way to test this function is by passing it data over a socket. The way it has to work is first read the headers and no more. Interpret the headers. If there's `Content-Length`, read *exactly* that number of bytes, no more, no less (this is not theoretical: if you don't do this, GitHub's requests won't work!). So it has to read from a stream, not a string.

ltratt:
Squashed.

src/httpserver.rs
vext01: sure, but you could split the tcp reading from the logic. The logic could accept a tuple of `headers: &str, data: &str`, then you can test the inner.

ltratt: It's worse than that: you *have* to parse the headers before you try reading the body from the network. [This was news to me, FWIW, and cost me a couple of hours debugging yesterday!]

ltratt:
@vext01 I *think* the problem might have been simple that docker doesn't set `$SHELL` (which, on reflection, is a reasonable thing for it not set). https://github.com/softdevteam/snare/pull/122/commits/18c887d95ee99b52e8d8b1c0d08278c3f4ea4641 fixes that (and I think should remain a separate commit). This works for me locally now, and is probably worth trying to merge as-is if you're OK with it.

PR #123, opened 2023-12-18
Test authentication.

ltratt:
In a sense these tests are the two most important missing tests, because they're what give confidence that snare is only executing commands when the request is properly authenticated.

### DIFF
--- a/tests/basic.rs
+++ b/tests/basic.rs
@@ -177,3 +177,128 @@ github {{
         exit_success,
     )
 }
+
+#[test]
+fn bad_sha256() -> Result<(), Box<dyn Error>> {
+    // Takes the SHA256 from [full_request], alters just the last digit, and checks that snare
+    // doesn't execute any commands (so, by proxy, we assume that snare doesn't authenticate the
+    // request).
+
+    let td = Builder::new().tempdir_in(env!("CARGO_TARGET_TMPDIR"))?;
+    let mut tp = td.path().to_owned();
+    tp.push("t");
+    let tps = tp.as_path().to_str().unwrap();
+    let req = r#"POST /payload HTTP/1.1
+Host: 127.0.0.1:28084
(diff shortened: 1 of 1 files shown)


tests/basic.rs
```diff
@@ -177,3 +177,128 @@ github {{
...
+        exit_success,
+    )
+}
+
+#[test]
+fn wrong_secret() -> Result<(), Box<dyn Error>> {
+    // Takes the example from [full_request], alters the client-side secret, and checks that this
+    // causes snare not execute any commands (so, by proxy, we assume that authentication failed).
+
+    let td = Builder::new().tempdir_in(env!("CARGO_TARGET_TMPDIR"))?;
```
vext01, line 248: Let's lift the common setup into a function?

tests/basic.rs
```diff
@@ -177,3 +177,128 @@ github {{
...
+        &format!(
+            r#"listen = "127.0.0.1:28085";
+github {{
+  match ".*" {{
+    cmd = "touch {tps}";
+    secret = "secretsecret";
+  }}
+}}"#
+        ),
+        move |_| {
```
vext01, line 222: I'm struggling to find the bit where you check that it failed.

tests/basic.rs
ltratt: The next PR will do just that :)

tests/basic.rs
ltratt: We (indirectly, which is the best we can do) check that the action wasn't executed (i.e. the named temporary file wasn't created) in https://github.com/softdevteam/snare/pull/123/files#diff-e41c063cd26a1181211ca22e91577e32d5067e01cf8f6341954e708842bf5f07R296.

vext01: I see.

PR #124, opened 2023-12-19
Fixes suggested by Clippy.

ltratt:


ltratt:
Closing because Clippy nightly has even more suggestions, so I might as reraise a combined PR.

PR #127, opened 2023-12-20
Further reorganise the tests

ltratt:
This PR heavily reorganises the new tests, including factoring out more common stuff. It also enables us to add a way of testing "snare failed", so we can readd in a "check that a bad config causes snare to exit with an error".


tests/common/mod.rs
```diff
@@ -126,3 +96,41 @@ where
         Ok(None) => Err("timeout waiting for snare to exit".into()),
     }
 }
+
+#[allow(dead_code)]
```
vext01, line 100: I'm guessing this can't be `#[cfg(test)]` due to the way you test?

vext01:
Just one query.

tests/common/mod.rs
ltratt: Because each `.rs` file that `mod`s this file is a crate, it means that it can be seen as "unused" in one `test/*.rs` file and not another. It's a bit messy, but I don't know of another way of avoiding meaningless warnings.

vext01:
There's a mq bug where a succeeding check links to the wrong build.

Could be that?

If you rummage in the buildbot Web UI you might be able to find the actual build?

ltratt:
Aha, yes, you're right! https://ci.soft-dev.org/#/builders/1/builds/3213

PR #128, opened 2023-12-22
Add more tests

ltratt:
This PR adds more tests to snare. While there is always more that we could do, this feels like it gets us to a "minimum acceptable" level of testing. Better late than never!

### DIFF
--- a/tests/queue.rs
+++ b/tests/queue.rs
@@ -1,0 +1,103 @@
+use std::{error::Error, fs::read_dir, thread::sleep};
+use tempfile::Builder;
+
+mod common;
+use common::run_success;
+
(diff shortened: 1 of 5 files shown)


tests/queue.rs
```diff
@@ -0,0 +1,103 @@
+use std::{error::Error, fs::read_dir, thread::sleep};
+use tempfile::Builder;
+
+mod common;
+use common::run_success;
+
+// Note that `sleep_ms` has to be a fairly high value as we really hope that snare has finished
```
vext01, line 7: sleep_ms -> sleep_s?

ltratt: Oops! Fixed in 9d08540.

vext01:
Please squash

ltratt:
Squashed.

PR #129, opened 2023-12-23
Finer grained logging

ltratt:
Needs #128 to be merged first.

This PR moves snare to a more fine-grained logging framework, and then makes use of that finer-grained logging to avoid bombarding the user with irrelevant stuff in the common case. For example, before this PR, invalid HTTP requests (caused, almost certainly, by spammers) generated logging, which isn't very useful. After this PR one only sees that if `-vvvv` is specified. At the default level of verbosity, however, more genuine errors (e.g. requests that don't authenticate) are logged.


ltratt:
OK for me to rebase?

vext01:
Go ahead.

ltratt:
Squashed.

PR #130, opened 2024-01-08
Use a Makefile that does a more portable installation.

ltratt:
This is based on recent changes to pizauth's Makefile. It respects `$DESTDIR` and also installs man pages in the correct (different!) location on both BSD and Linux. The downside is that the Makefile is now GNU make only.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,19 +1,29 @@
 PREFIX ?= /usr/local
-MAN_PREFIX ?= ${PREFIX}/man
+BINDIR ?= ${PREFIX}/bin
+LIBDIR ?= ${PREFIX}/lib
+SHAREDIR ?= ${PREFIX}/share
+EXAMPLESDIR ?= ${SHAREDIR}/examples
+
+MANDIR.${PREFIX} = ${PREFIX}/share/man
+MANDIR./usr/local = /usr/local/man
+MANDIR. = /usr/share/man
+MANDIR ?= ${MANDIR.${PREFIX}}
+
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,19 +1,29 @@
...
-MAN_PREFIX ?= ${PREFIX}/man
+BINDIR ?= ${PREFIX}/bin
+LIBDIR ?= ${PREFIX}/lib
+SHAREDIR ?= ${PREFIX}/share
+EXAMPLESDIR ?= ${SHAREDIR}/examples
+
+MANDIR.${PREFIX} = ${PREFIX}/share/man
+MANDIR./usr/local = /usr/local/man
+MANDIR. = /usr/share/man
+MANDIR ?= ${MANDIR.${PREFIX}}
```
vext01, line 10: Your GNU foo is better than mine. How does this alter the man directory based on platform? Does it rely on one of the platforms passing `${PREFIX}` at build time?

ltratt: It was a suggestion by someone else -- I wouldn't have thought of this! Basically if you pass `PREFIX=/usr/local` then `MANDIR` is set to `/usr/local/man`, otherwise it's set to `/usr/share/man`. It's very cunning!

vext01: I bet there's a platform where `PREFIX=/usr/local/`, but the manuals go into `/usr/local/share/man`, but let's not worry about it until we see it.

I thought FreeBSD did it this way, but I just check and I was wrong.

PR #133, opened 2024-02-05
Rethink logging

ltratt:
In https://github.com/softdevteam/snare/commit/45666ba0c1cbb0d1bc4a019169197fef1fbf95d4 we moved to the `log` crate and friends --- but that's just bitten us with `atty` as a dependency which has a security advisory.

This PR removes `log` and back to our own (small) custom logging (that deals with stderr and syslog) (https://github.com/softdevteam/snare/commit/a61aec7142d2de140c621db99aa2faa6a9e75273). As a bonus I finally added the ability to see which requests are coming in (with `-vv`) (https://github.com/softdevteam/snare/commit/eba8bac6ae767a328bb5fd99ebfd8d1258a222e4). Both commits contain a bit more detail. Lightly tested on b13 and my main server and working fine.


vext01:
Are you sure you want to do this? The `log` crate is pervasive across the rust ecosystem. Someone must be working on a fix?

ltratt:
Maybe they are --- but this change also reduces the binary size by almost 10%, so it's a win in other ways.

vext01:
It's up to you. I'm not sure I'd have gone my own way if it were mine. People know how `log` works, disk is cheap and you wouldn't have to have all that extra code.

By the way, your vuln doesn't come from `log` at all, it comes from `stderrlog`:
```
atty 0.2.14
└── stderrlog 0.5.4
    └── snare 0.4.9
```

You could still use one of the standard logging libs if you wanted, for example `env_logger` which has no known vulns.

If you say merge, I will, but I'd just use `env_logger` and be done with it :P

ltratt:
Honestly, for a daemon, I think the lightweight logging in this PR is probably for the best. I should have stuck with it in the first place!

PR #134, opened 2024-03-14
Unbreak `ping` support.

ltratt:
This didn't return an HTTP result, so github was quite reasonably upset! Add a test so this problem can't recur.


vext01:
Looks OK code-wise. Don't know enough about this stuff to say anything profound, so will have to trust you on this :)

PR #138, opened 2024-07-21
Update README.md with Content Type instructions for github integration.

mweitzel:
Update docs to reflect that snare supports the Content-type `application/x-www-form-urlencoded`, not `application/json`.

This bit me while setting it up locally. I kept getting _"Payload does not start with 'payload='"_ error messages. When inspecting the test code I [thought it only supported `application/json`](https://github.com/softdevteam/snare/blob/4f94c26bd23b03f9422d610644c449896b9696c5/tests/actions.rs#L56), even though the payload isn't actually json.

Would you want a PR to supporting both `x-www-form-url-encoded` and `json`? It looks like branching on the `Content-Type` header [here](https://github.com/softdevteam/snare/blob/4f94c26bd23b03f9422d610644c449896b9696c5/src/httpserver.rs#L119-L131) would work well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,12 +159,14 @@ about other parts of the file system.
 
 `snare` runs an HTTP server which GitHub can send webhook requests to.
 Configuring a webhook for a given GitHub repository is relatively simple: go to
-that repository, then `Settings > Webhooks > Add webhook`. For `payload`,
-specify `http://yourmachine.com:port/`, specify a `secret` (which you will then
-reuse as the `secret` in `snare.conf`) and then choose which events you wish
-GitHub to deliver. For example, the default `Just the push event` works well
-with the email diff sending per-repo program above, but you can specify
-whichever events you wish.
+that repository, then `Settings > Webhooks > Add webhook`. 
+- For `payload`, specify `http://yourmachine.com:port/`
+- For `Content type`, select `application/x-www-form-urlencoded`
+- Specify a `secret` (which you will then reuse as the `secret` in `snare.conf`)
+
+And then choose which events you wish GitHub to deliver. For example, the default
+`Just the push event` works well with the email diff sending per-repo program above,
+but you can specify whichever events you wish.
 
 Once you have set up your webhook, GitHub will automatically send a test "ping"
 message. `snare` acknowledges pings (i.e. confirms to GitHub that it is


ltratt:
I'm not sure I ever knew that `x-www-form-url-encoded` and `json` were treated differently by github, and if I did know, I've forgotten why! Do you have an example of what the different queries that snare might receive for each look like?

Background: the test suite was created some time after snare was created, and from memory I used an example on github as a template. I probably didn't even realise `x-www-form-url-encoded` and `json` existed as separate things. I wonder if I added the `X-` headers myself? The fact that the test suite works suggests that snare is mostly agnostic as to this difference, but perhaps this is based on a completely false understanding on my part.

Ultimately, though, if we can support both, I'd welcome that: anything that reduces friction for users seems look like a good thing to me!

mweitzel:
I do; here's an example:
- [this repository](github.com/mweitzel/junk) has webhooks configured to "give me all events"
- it is set up twice, for both url-encoded and json content-types, each with the secret "asdf"
- to deliver webhooks to webhook.site service (you can see events here https://webhook.site/#!/view/3cc71448-b8f7-49d1-a286-c0d1b0102100/d2acf01c-ce99-48df-8825-c8b27614b38c/1)

For each event, you can see two entries, one from each webhook.

This service has very helpful "copy as curl" feature, where I edit the curl url from `webhook.site/uuid` to `localhost:8011` to test manually against a local instance of snare. (As well as the "HAR JSON" format I've found useful for automated testing)

_Aside: in curl, the `-d xxxxxx` flag specifies the data of the POST body_

The `application/x-www-form-urlencoded` curl looks like:
```
curl -X 'POST' 'https://webhook.site/3cc71448-b8f7-49d1-a286-c0d1b0102100' -H 'connection: close' -H 'x-hub-signature-256: sha256=5b0b1a6d8eca7232d785627fbc714a03e3b63e08f92ec70a67ee7c13cc4b1369' -H 'x-hub-signature: sha1=8e427770590c30433e18c1767b25dd831e2bd081' -H 'x-github-hook-installation-target-type: repository' -H 'x-github-hook-installation-target-id: 310045141' -H 'x-github-hook-id: 491392686' -H 'x-github-event: pull_request' -H 'x-github-delivery: 83ea3ad0-477b-11ef-975e-b59c694765a2' -H 'content-type: application/x-www-form-urlencoded' -H 'accept: */*' -H 'content-length: 29630' -H 'user-agent: GitHub-Hookshot/74908eb' -H 'host: webhook.site' \
-d $'payload=%7B%22action%22%3A%22closed%22%2C%22number%22%3A3%2C%22pull_request...
```
and the `application/json` curl looks like:
```
curl -X 'POST' 'https://webhook.site/3cc71448-b8f7-49d1-a286-c0d1b0102100' -H 'connection: close' -H 'x-hub-signature-256: sha256=1e68a74005e80d052430796ea6ec42aad2e3cd0febd94cd451572b612b6c460e' -H 'x-hub-signature: sha1=71da611cd2eb9588bf3e14b6921f280600fefd54' -H 'x-github-hook-installation-target-type: repository' -H 'x-github-hook-installation-target-id: 310045141' -H 'x-github-hook-id: 284243233' -H 'x-github-event: pull_request' -H 'x-github-delivery: 83ea3ad0-477b-11ef-975e-b59c694765a2' -H 'content-type: application/json' -H 'accept: */*' -H 'content-length: 20838' -H 'user-agent: GitHub-Hookshot/74908eb' -H 'host: webhook.site' \
-d $'{"action":"closed","number":3,"pull_request":...
```

The main differences are the `Content-Type` and `x-hub-signature*` headers. I *think* the signature is purely off of the payload, not the entire request. It looks like url-encoded version is just a form with a single value (`payload=`) with content that matches the json payload.

mweitzel:
Closing in favor of #139

PR #139, opened 2024-07-22
Handle both JSON and urlencoded requests.

ltratt:
I singularly failed to realise that GitHub requests can come with a Content-Type of JSON or urlencoded: previously snare only handled the latter. Handling both automatically is fairly easily, but doing so highlighted that the tests incorrectly said they had JSON content-type but were actually urlencoded. Since JSON content-type is easier to understand, the tests now consistently use that, except for a (new) test to ensure that we really handle urlencoded correctly.

@mweitzel Does this work for your repo(s) on both JSON and urlencoded webhooks?

### DIFF
--- a/src/httpserver.rs
+++ b/src/httpserver.rs
@@ -116,19 +116,42 @@ fn request(snare: &Arc<Snare>, mut stream: TcpStream) {
         None => None,
     };
 
-    if !body.starts_with("payload=".as_bytes()) {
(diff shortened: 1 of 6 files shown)


ltratt:
Thanks for testing! It was a good bug report which made the fix obvious -- my favourite :)

vext01:
LGTM

PR #141, opened 2025-02-28
Housekeeping

ltratt:
This PR updates snare's dependencies and fixes a few Clippy complaints.


ltratt:
Force pushed `cargo fmt` update.

ltratt:
It turns out that quite a bit has changed since we last run `cargo deny check`! Pushed commits to address that (they should remain separate IMHO).

